PR #20, opened 2022-05-03
Idea of feature for streamers : Set pokemon's experience for evolution checking

GB127:
So I had this idea while watching some streamer. Sometimes we "just want to know" what is the evolution of a pokemon we just lost... Sometimes the level is close to grind the experience. Sometimes it is not.

This addition will set the experience of the pokemon on slot 0 to 60 000 experience. In the fluctuating category, it would be level 39ish if my calculations are correct.

To use this feature:
=> Right click on the emu window (This command needs to be changed. Perhaps a new button? I just wanted a prototype working... Not to mention that lua scripting is not my forte)
=> In the game,  STORE the pokemon in a computer (Do not open the pokemon summary via the pause menu yet!)
=> Withdraw back the pokemon
=> The pokemon should be at a higher level.
=> Evolve the pokemon via grinding against wild pokemons.

In summary, the addition edits the experience stored in the data of the pokemon. The checksum is also edited so we don't get a badd egg. this works because the game only keeps the experience data when storing in a computer. The level is just a number the game calculates with the experience data and it is not kept after storing in the computer.


GB127:
Version used for this : Firered (USA Europe) Rev 1

The pokemon offsets will maybe need to be changed to support all versions.

besteon:
I think a better option for this would just be to read/display the evolution mon data on the tracker itself. I see no reason that you would need to go through the process of storing the mon in the PC and such. We want the features to be as seemless and easy to use as possible.

And while this could be a useful feature, I would want to implement it in such a way that the feature is disabled until you white out so that it couldn't be abused. Right now you can always look at the randomizer logs to view the would-be evolution mon.

I encourage you to make a feature request for this idea.

PR #38, opened 2022-05-22
Add notes feature and make reset combination always clear data

schwjm:
- Allows setting text notes using the empty space at the bottom of the tracker
- Allows resetting the tracker without using the `ROMS_FOLDER`/`LoadNext` feature
image


bdjeffyp:
I really think that this is a great use of the space and leveraging BizHawk's Lua forms.
A couple of things to note:

1. You can overflow the region... 
image
Is it possible to get the text cut off by the box region?
2. Perhaps adding a label to the box, like "Notes:" or "Note:", can help users know the intent of the box. In any case, an update to the Readme.md file is needed to describe the feature and how to use it.
3. I typed this long string: "My uncle, who works at Nintendo, found that there was a secret to catching Mew. You have to do a bunch of ridiculous shit and then, BOOM, you got yourself a Mew! Incredible!" What issues are there if people type strings that are too long? I recommend making a character limit for the string.
4. Add something to address non-ASCII characters. I'd hate to see what unicode/emoji can do to this thing. 😅

besteon:
> I really think that this is a great use of the space and leveraging BizHawk's Lua forms. A couple of things to note:
> 
>     1. You can overflow the region...
>        image
>        Is it possible to get the text cut off by the box region?
> 
>     2. Perhaps adding a label to the box, like "Notes:" or "Note:", can help users know the intent of the box. In any case, an update to the Readme.md file is needed to describe the feature and how to use it.
> 
>     3. I typed this long string: "My uncle, who works at Nintendo, found that there was a secret to catching Mew. You have to do a bunch of ridiculous shit and then, BOOM, you got yourself a Mew! Incredible!" What issues are there if people type strings that are too long? I recommend making a character limit for the string. This would then address issue 1 above.
> 
>     4. Add something to address non-ASCII characters. I'd hate to see what unicode/emoji can do to this thing. 😅

I think a box label might restrict the character limit too much since the label and the note would have to fit. However, I think a placeholder such as [click for note] or something similar would be useful when there is not currently a note set.

besteon (changes requested):
Add placeholder value and limit note length.

schwjm:
I agree with your suggestions but I am at a loss as to how to implement most of them. This is my first foray into Lua and Bizhawk scripting so I may be missing something obvious. The text functions available don't seem to offer anything to determine the width of a string given a certain font. This may be available with a third-party library, but it seems overkill. The forms API itself is also very limited--there does not seem to be a way to limit input here either, so at best it would have to cut things off after the fact?

Of course, a fixed width font could be used, but that would not look as good. My thought was that doing no check ensures the maximum amount of space can be utilized at the risk that someone puts in something too long. Seems harmless enough, though.

Here is a sample of using an "edit" icon instead of making text, for a more minimalist approach. I can add text still if you prefer (but would prefer it is not high contrast because it seems like it would draw too much attention).

image
image

> 4\. Add something to address non-ASCII characters. I'd hate to see what unicode/emoji can do to this thing. 😅

Could you elaborate on this concern? I think it falls under the same issue, a user can avoid doing something that would not work and if it doesn't work, it's fairly harmless. The system is at the mercy of what that font at 9pt can render well.

> 2\. Perhaps adding a label to the box, like "Notes:" or "Note:", can help users know the intent of the box. In any case, an update to the Readme.md file is needed to describe the feature and how to use it.

I've added something to README.txt, but there isn't really much about "how to use" other than how to load it. I'll keep it brief and more can be done here probably. README.md has less in this regard so I welcome some guidance on how to start there.

schwjm:
> The placeholder icon is amazing and I think that is the way to go here. My only suggestion is that the icon be removed when a note does exist to fill the space.

I was torn on whether to make it disappear or not, either way works for me.

> My concern with strings too long and unicode characters stems from concerns for buffer overflows and such. I don't know what BizHawk does concerning it's Lua engine. Perhaps it is in a sandbox and just shuts the engine down when an issue like that occurs. Perhaps not. Maybe the worst thing that happens is that BizHawk just crashes and the enterprising player who tried to break things looses their progress. Ultimately, we should be placing safeguards in place to protect the users from shooting themselves in the foot and breaking their game. Telling a user to avoid doing something never works. You have to prevent it. There is regex out there for scrubbing out non-ASCII characters which you can implement when the user submits the note.

I don't know of any character that would be problematic to play back via `drawText`, so I don't see an action to take on non-ASCII characters at this time. Any solution here is still going to be "something unexpected" because either something draws funny, or user-entered data is deleted with no feedback (or primitive feedback at best). I can prove out that stranger edge cases do or don't crash Bizhawk and handle accordingly if anything comes up.

It is fair to store only the first X characters, just in case, to stop extraordinary abuse from causing memory issues.

> You can also trim any characters past some static size, say 25 or 30 characters in length. This meets the intent of the requirement that the text does not spill over out of the box while also not entering too many characters that the user can potentially break the game.
What I'm arguing is that being able to always use the full width of the box as a requirement trumps the requirement that things always fit in the box. Because this is not a fixed-width font, the more Is and ls and other narrow characters, the more likely you'll be cut off significantly before the edge if a worst-case limit is implemented.

I think there is some value in letting people see "more" than what the tracker shows (clicking the Note shows you exactly what you typed whether it fits on the game screen or not).

Therefore, a good middle ground here might be to draw over any spillover text (rerender in the area where spillover could happen), in lieu of a way to limit the width via `drawText`. This stops the unsightliness/silliness of text spilling out. Then the character limit can be set at something like 50-100 which is more than you'll ever need to fill up the box, and probably more than anyone would ever use realistically.

> For the READMEs, we are just looking for a basic thing there. Doesn't need to be too detailed. Perhaps just a bullet point after the bullets on the Settings.ini file and Quick Loading Seeds.

See 8616458 for my attempt.


As an aside, @besteon -- can I get a ruling on tabs or spaces? This repo could use a .editorconfig to set the standard (and probably .gitattributes too) as I've already run into issues keeping things consistent (and they're not consistent between files).

bdjeffyp:
This all sounds good to me!

Don't sweat the README stuff, after all. I took care of updating the README.md in another branch that will be joined soon. It includes a small blurb about the Notes feature in a "coming soon" section.

besteon:
I like the idea of a 50 character limit. I don't think you need more than that for each mon. If you want 100, we could also do that. I don't see a need to limit it too much while we haven't witnessed any save state memory issues yet.

I also like removing the note icon when a note is present. Perhaps we should expand the note clicking area to the entire row when the icon isn't present (so clicking anywhere on the note will edit it).

Maybe just use string.byte() on each character to make sure it is within standard ascii range? Might be the simplest method.
https://www.gammon.com.au/scripts/doc.php?lua=string.byte

schwjm:
> I like the idea of a 50 character limit. I don't think you need more than that for each mon. If you want 100, we could also do that. I don't see a need to limit it too much while we haven't witnessed any save state memory issues yet. When displaying a note that is over the visible limit, I would either cut it off or append "..." to signal to the user that it can be clicked for the rest of the note.

Since we can only blindly redraw over areas where spillover could appear, "..." isn't feasible, but I would prefer that if it could be detected. For now, the text is cut off by doing some additional drawing.

image


> I also like removing the note icon when a note is present. Perhaps we should expand the note clicking area to the entire row when the icon isn't present (so clicking anywhere on the note will edit it).

The entire box is the collision area, the icon is just decorative.



> Maybe just use string.byte() on each character to make sure it is within standard ascii range? Might be the simplest method.

So when testing things outside of ASCII range, I discovered that perhaps Bizhawk `forms` does not support it to begin with. From cursory reading, Lua is primitive enough to support it / or not as long as you know what you're doing. However, when entering in some random emojis and Japanese characters, all of the corresponding bytes were converted to `?`. So I don't think there's anything to do here, they don't work and we cannot detect that someone tried something that can't work.



> Also, use tabs.

Noted. I'll submit a .editorconfig PR in the future along with normalization if I contribute further. For now I have kept things consistent in each file as best as I could.

I believe this is ready for another review pass. I noticed the form didn't go away automatically when the tracker script exited, so that's cleaned up now.

bdjeffyp (approved):

besteon (approved):
Text limit accounted for :)

PR #40, opened 2022-05-22
Small formatting to better align the values.

GB127:
However. There is a problem : The 0s are ugly. I tried a couple of time. The fix is to remove the 0s in the formatting.

The current issue that forced me to use zeros as "proof of concept" is that the spaces' width aren't equal to the space of the digits in the font used. To replace the zeroes, we would need to change the font used for the numbers. Before I go ahead and try to edit the font (I have no idea how),  I want to confirm if that is something you want incorporated before I lose my time trying to figure this out.

I wrote this because, honestly, I prefer a aligned display. I dislike having a display like this:
```
 95
 90
 100
```
I prefer something like:
```
  95
  90
 100
```
In case this isn't shown properly on a different browser, all the hundreds should be aligned, all the tens are aligned, and all the units are aligned.


GB127:
I will also fix the code formatting after I finish my modification if this kind of modification is welcome. Don't worry about the mess that I made with all the new lines and all.

ironmon_tracker/Drawing.lua
```diff
@@ -424,21 +424,31 @@ function Drawing.DrawTracker(monToDraw, monIsEnemy, targetMon)
 	local ppOffset = 82
 	Drawing.drawText(GraphicConstants.SCREEN_WIDTH + ppOffset, moveStartY - moveTableHeaderHeightDiff, "PP")
 	for moveIndex = 1, 4, 1 do
-		Drawing.drawText(GraphicConstants.SCREEN_WIDTH + ppOffset, moveStartY + (distanceBetweenMoves * (moveIndex - 1)), Utils.inlineIf(monIsEnemy or moves[moveIndex].pp == NOPP, moves[moveIndex].pp, Utils.getbits(monToDraw.pp, (moveIndex - 1) * 8, 8)))
+		Drawing.drawText(
```
GB127, line 427: Example of mess I made that I will fix if I get the green light.

bdjeffyp:
I hear you on the text alignment. Can you provide screenshots of what your change looks like in the tracker? I noticed that this only covers the moves section of the tracker. What about the stats?

bdjeffyp:
I'm also not certain what you mean by the first line in initial comment: "However. There is a problem : The 0s are ugly. I tried a couple of time. The fix is to remove the 0s in the formatting." Can you describe further what you mean here?

GB127:
I wasn't bothered for the stats. But I can adjust those as well.

 "However. There is a problem : The 0s are ugly. I tried a couple of time. The fix is to remove the 0s in the formatting." Can you describe further what you mean here?

What I mean here is that I tried with formatting with spaces instead of zeros. But because spaces aren't as wide as the zeroes, I couldn't get a decent look.

Currently it only edit the PPs, move power and move accuracy.

GB127:
Here is a page on monospaced font : https://en.wikipedia.org/wiki/List_of_monospaced_typefaces

The font used for the digits is not monospaced. That is the current issue.

bdjeffyp:
The numbers in this font are definitely fixed width. All digits are four pixels wide and use a leading empty pixel column if only 3 pixels wide (in the case of "1")

bdjeffyp:
Example screenshot:
image
These digits are fixed width

GB127:
But the spaces aren't it seems. Unless I misunderstand something from the lua string formatting.

From what I understand from lua scripting, the fix to replce the zeros to spaces is to simply remove the 0.
"%02s" => "%2s".

I was able to add a lot of spaces and everything moved as expected (I tried wiht something like "%8s"). But even if I put "%3s", the numbers don't align as it should be. I'm not sure what is the issue if the font used is monospaced.

GB127:
image

PP is "2%s".
Accuracy and Power are "%3s".

bdjeffyp:
@GB127, I want to personally apologize because I have misread what you are stating multiple times. I believe I understand the problem now. Just to make certain I am not messing this up again, I'll restate to make sure I have this right.

You want to align numbers, but you can't because the "space" character itself is not at a fixed width, because the whole font is a variable width font. I was overly concerned about the numbers themselves because I thought that was what you were referring to specifically. Again, I apologize. 😅

So, here is my suggestion then:
- Make a new function in the Drawing object, like `drawText`, for numbers specifically. This function needs to behave similarly to `drawText` (has two calls to `gui.drawText` for drawing the shadow and then the text) but it can use a monospace font.
- Name the new function `drawNumbers`. This will call out the fact that the function is explicitly for rendering numbers in the tracker (and any aligning spaces).
- The most important part: In the two calls to `gui.drawText` within your `drawNumbers` function, you need to choose a font family that is monospaced, but ALSO is similar to the current font. I recommend choosing a few of them and taking screenshots of the appearance so that Besteon can make the ultimate choice on which to go with, or if they'd rather stick with the current appearance.

I hope that this gives you a clearer path forward to accomplish something that can make this tracker project more aesthetically pleasing to the eye. I'm sorry again for the earlier run around and thank you for taking the effort to make this project better for everyone!

GB127:
Noted. Will look into it today or tomorrow.

Do not worry about the misunderstanding. :)

GB127:
-- Tested with 3 of the 4 monoscripts according to this site: https://www.autoitscript.com/autoit3/docs/appendix/fonts.htm
        -- In my opinion, Ludida Console gave the best result. Consolas gave good results too.
        -- I'm not sure what other fonts are accessible...

GB127:
Here are the outputs:
Consolas
consolas
Courier New
courier new
Lucida Console
Lucida console

GB127:
I have updated the code in my repo, but I don't know how to update this pull request...

GB127:
Currently fixing the conflicts. Hold on...

GB127:
Just got an idea that would allow me to avoid editing the font. I'll implement the idea tomorrow.

GB127:
image

.gitignore
```diff
@@ -0,0 +1 @@
+calculate_stat_prototyp.py
```
bdjeffyp, line 1: Please delete this file and recommit

bdjeffyp (changes requested):
Aside from deleting the .gitignore, a new Settings.ini entry should also be included to give users the option to right justify the numbers or use the original style.

.gitignore
GB127: Oops. My bad.

GB127:
> Aside from deleting the .gitignore, a new Settings.ini entry should also be included to give users the option to right justify the numbers or use the original style.

I have no idea how to do this...

.gitignore
GB127: I deleted my gitignore and recommited.

GB127:
Here what I have done :
-> New option in settings
-> Check a bool with the setting. If justified, add a spacing. Else, add no spacing.

bdjeffyp (approved):

PR #43, opened 2022-05-23
Add Features section to README

bdjeffyp:
Attempting to clean up README.md more, giving a place for all of the features of the track. Hopefully, this will help users know more about what the tracker can do for them right up front. This will also help future collaborators to have a good place to add documentation on their features.

Oh yeah, I also added more links and references to BizHawk needing to be version 2.8 or higher... I held back on my snark... 😅

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -9,6 +9,15 @@ PLACEHOLDER = "---" -- TODO: Consider moving into a better global constant locat
 
 print("Ironmon-Tracker v" .. TRACKER_VERSION)
 
+-- Check the version of BizHawk that is running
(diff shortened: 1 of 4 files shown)


bdjeffyp:
Added a version check near the beginning of the script to ensure that users are using version 2.8. If not, the script will end immediately. This will resolve a LOT of errors that people keep bringing up in Discord!

Note: Currently, it checks the output of client.getversion() explicitly as a string compare. So if there is a version upgrade, we would need to add the strings to the `if` check, which is not ideal or scalable. I'm sure there is a better alternative to how I did this, but this works for now.

besteon (approved):

PR #45, opened 2022-05-24
Version 0.2.2 release

bdjeffyp:
Iterates the version number in preparation for release
Also adds the `forms` object for BizHawk to the globals for VSCode settings and an .editorconfig to enforce tabs as the spacing style.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,5 +2,4 @@ root = true
 
 [*]
 indent_style = tab
-end_of_line = crlf
 trim_trailing_whitespace = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*]
+indent_style = tab
+end_of_line = crlf
+trim_trailing_whitespace = true
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -2,7 +2,7 @@
 -- Created by besteon, based on the PokemonBizhawkLua project by MKDasher
 
 -- The latest version of the tracker. Should be updated with each PR.
-TRACKER_VERSION = "0.2.11"
+TRACKER_VERSION = "0.2.2"
 
 -- A frequently used placeholder when a data field is not applicable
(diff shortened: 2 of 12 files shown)


Ironmon-Tracker.lua
```diff
@@ -2,7 +2,7 @@
 -- Created by besteon, based on the PokemonBizhawkLua project by MKDasher
 
 -- The latest version of the tracker. Should be updated with each PR.
-TRACKER_VERSION = "0.2.11"
+TRACKER_VERSION = "0.2.2"
```
schwjm, line 5: Consider [Semantic Versioning](https://semver.org/) where each component is a separate number, as opposed to some hybrid decimal/fractional number. Therefore the "12th patch" (0.2.12) is next, and the "2nd patch" would come before the 11th.

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+
+[*]
+indent_style = tab
+end_of_line = crlf
```
schwjm, line 5: Instead of enforcing crlf which is a bit Windows-centric, consider a [.gitattributes](https://git-scm.com/docs/gitattributes) with 
```
* text=auto
```
which tells Git that anything it can detect as text, it will fix line endings according to platform and settings.

schwjm: Also, certain files like Tracker.lua are currently using spaces, consider running autoformatting on all files.

Ironmon-Tracker.lua
bdjeffyp: Crap, I had written a bunch of stuff here, clicked "Resolve" and it didn't leave the comment... I'm too used to Azure DevOps. 🤦‍♂️

Anyway, what I had said previously was that the last version should actually have been 0.2.2 and this would then be 0.2.3, if we follow semantic versioning. Calling it 0.2.11 was my mistake as I was trying to say it was just a very minor update to 0.2.1...

I think we can continue on with semantic versioning from here, but I'll leave this as 0.2.2 for now.

PR #53, opened 2022-05-29
Attempt to add low kick calculating for convenience

Brian0255:
Since I see people often having to look up pokemon weights for Low Kick, I thought it would be a good idea to manually add all the pokemon weights to PokemonData and calculate Low Kick's damage on the fly automatically (by adding a function to the Utils file). Tested and works for both your mon and the enemy mon. Will show "WT" outside of battle.

### DIFF
--- a/ironmon_tracker/Utils.lua
+++ b/ironmon_tracker/Utils.lua
@@ -40,7 +40,7 @@ function Utils.isSTAB(move, pkmnData)
 	return false
 end
 
-function Utils.calculateLowKickDamage(weight)
+function Utils.calculateWeightBasedDamage(weight)
 	if weight < 10.0 then
 		return "20"
 	elseif weight < 25.0 then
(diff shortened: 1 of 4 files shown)


bdjeffyp (changes requested):
Thanks so much for this work! Since you beat Fellshadow to the punch in this submission, we will take your PR. One thing I will ask you is that you add a new setting to the `Settings.ini` file to give people the option to use this feature or not. Make sure that your new setting is documented with a comment to describe what it does and then add a check for it to be set to `true` in the `Drawing.lua` file.

ironmon_tracker/Utils.lua
```diff
@@ -39,3 +39,19 @@ function Utils.isSTAB(move, pkmnData)
 	end
 	return false
 end
+
+function Utils.calculateLowKickDamage(weight)
```
bdjeffyp, line 43: Please change the name of this function to something more generic, since it can also be used with Grass Knot in Gen 4.

bdjeffyp (approved):

PR #54, opened 2022-05-29
Add low kick power calculation

Fellshadow:
Heyo, I saw pie constantly looking up weights to calculate low kick power and got curious about adding it in

From what i've tested so far seems to work, and should in theory work for enemy mons using low kick on you (showing the power your weight gives them) but the latter part there is hard to test. It still shows the default value "WT" outside of battle.

I'm aware of a similar pull request made (#53 ) so seems they did it at the same time lol but still adding this for your consideration between which to use

Calculation and data were taken from serebii and should also work for grass knot should gen 4 support be added sometime in the future


bdjeffyp:
Thanks for the submission, Fellshadow. Since OnlySpaghettiCode beat you to the punch, we are going to go with theirs instead. I really appreciate the effort and time you put into this and helping to make this tracker better for everyone.

Fellshadow:
no worries, I felt that'd be the case but figured i'd shoot my shot anyways haha 

who knows, maybe another time i'll get there first ;) regardless keep up the great work guys!

PR #55, opened 2022-05-29
Hyper potion value fix

rcj001:
In gens 1-6, hyper potions heal for 200. I imagine you might want some logic to handle this if support grows to gen 7+, but for the currently supported games, 200 is the correct heal amount. 

https://pokemondb.net/item/hyper-potion


bdjeffyp (approved):
Nice catch! Thanks for submitting the fix!

rcj001:
No problem, thanks for all the work you and everyone else have put into the tracker! It has been a great addition to ironmon

PR #60, opened 2022-05-31
Fix Weight Move Calculation

Fellshadow:
hey, so fun fact i noticed a bug in the low kick calculation lol It only updates once. This is because the check was only looking for if the power was WT and then overwrote that power in the data itself, so in subsequent checks it would never be WT again

clip of the bug - https://clips.twitch.tv/MushyEnjoyablePeachAMPEnergy-brQVUTqCCSvyAe34

To fix this i re-used the calculation in the drawing side from the initial low kick calc i submitted but rewrote to use the utility function added by OnlySpaghettiCodes, which i've tested and know works (while switching back to WT after battle)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+indent_style = tab
+trim_trailing_whitespace = true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,15 +1,3 @@
 {
-    "Lua.diagnostics.globals": [
-        "event",
-        "emu",
-        "client",
-        "gui",
-        "userdata",
-        "gameinfo",
-        "bit",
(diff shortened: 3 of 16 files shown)


Brian0255:
Good catch (and dumb oversight by me), just make sure you're also checking that Settings.tracker.CALCULATE_WEIGHT_BASED_DAMAGE = true

Fellshadow:
oh yeah oops i must have forgotten to copy that over i'll get that in, one moment

Fellshadow:
ok that's in there now, thanks for catching that

Fellshadow:
small note, just tested with the setting off and the order of if statements i had broke it when the setting was set to false, fixed that

Fellshadow:
OK so first of all my apologies for all the small commits and stuff, in my slight haste to get a fix out i wound up being a bit too lax in my testing

i have now tested this in and out of battle, with and without the setting enabled, and it should NOW be working 😩

bdjeffyp:
Awesome! Thanks so much, both of you!

PR #64, opened 2022-06-02
Small idea : Automate estimation of opponent's stats

GB127:
Still a WIP, but want to make sure it's welcome... And to seek guidance to fully implement this.

In summary, I created a function that will accept a HP damage (in the future, will use also a boolean value for crits), and returns the two possibles ends of an attack stats of the opponent's pokemon.

Here is what it *would display* following an attack (Read the commentary on the code for the exact scenario).

The placement of the informaiton is still subject to moves. Perhaps create an option to switch from boxes to those estimations?
image

### DIFF
--- a/ironmon_tracker/Drawing.lua
+++ b/ironmon_tracker/Drawing.lua
@@ -1,30 +1,26 @@
 Drawing = {}
 
 
-
-
-
-
 local function calculate_opponent_attack(damage, crit)
     -- The idea behind this function is to automate the mathematics that
     -- you can do in theory with all the infos provided. The pokemon games's damage formula
     -- has been cracked down and you can find it on the web. This suggestion will display the range of
-    -- possible stats.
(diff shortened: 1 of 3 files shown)


GB127:
Maybe use an input box and a Special /  Physical checkbox? So that the user still have to calculate the HP lost?

valiant-code:
This is a level of automation that I think reaches too far into playing the game for you.

bdjeffyp:
I agree with valiant-code on this. Automation regarding opponent Pokémon must be avoided, otherwise the challenge is reduced. Predictions are incumbent on the player to make rather than the code.

bdjeffyp:
I agree with valiant-code on this. Automation regarding opponent Pokémon must be avoided, otherwise the challenge is reduced. Predictions are incumbent on the player to make rather than the code.

PR #65, opened 2022-06-02
Manual Hidden Power Feature

Brian0255:
So I thought it might be nice to let the player change the move type of Hidden Power manually (as opposed to it being calculated automatically somehow). This would allow a bit of convenience without being overly automated. My idea that I have implemented here is allowing the player to click the move's name to cycle through types, with the color, STAB effectiveness, and move effectiveness all changing dynamically.

The way I did this is to add a special Hidden Power button that is only drawn when the player has the move. If they have it, it gets drawn where the move would normally go, and responds to clicks accordingly. The tracker will store what the player has manually selected as Hidden Power's type, making it so that enemy mons with Hidden Power will still default to normal type.

I've tested this and it should work fully, so I would love to hear your thoughts on this feature or any changes I should make!

Here's a video as well:

https://user-images.githubusercontent.com/61294586/171597645-8c022839-71c5-481e-bf9e-65a8eb817bd4.mp4


Brian0255:
Thanks, great catch! I completely forgot about that and have updated it.

ironmon_tracker/Utils.lua
```diff
@@ -50,7 +54,11 @@ end
 
 function Utils.isSTAB(move, pkmnData)
 	for _, type in ipairs(pkmnData["type"]) do
-		if move["type"] == type then
+		local moveType = move.type
+		if move.name == "Hidden Power" and Tracker.Data.selectedPlayer == 1 then
+			moveType = Tracker.Data.currentHiddenPowerType
+		end
+		if moveType== type then
```
bdjeffyp, line 61: nit: put a space between `moveType` and `==`

ironmon_tracker/Program.lua
```diff
@@ -684,3 +684,14 @@ function Program.getBagStatusItems()
 
 	return statusItems
 end
+
+function Program.playerHasMove(moveName)
```
bdjeffyp, line 688: Can this be placed in Utils.lua? I feel like it is better scoped as a utility function rather than a main program function.

bdjeffyp (changes requested):
This is a great alternative to having to use the Notes feature while also being able to make use of the move effectiveness feature. Thanks for doing this, OnlySpaghettiCode! I only had a few things I'd like addressed before merging.

bdjeffyp (approved):

bdjeffyp:
Merged! Thanks again for another great addition!

PR #66, opened 2022-06-03
Modify settings in the tracker

bdjeffyp:
Players will now be able to modify all boolean settings and the Roms folder setting from the tracker itself!

image

By clicking on the gear icon, users are brought to the new Settings page.

image

Each toggle setting is built using the name of the setting itself, converted into proper casing for display on the page. I updated some of the setting names to be more descriptive of their function, since we don't have fancy tooltips or anything to state their function clearly. Also, clicking on the Roms folder option will bring up the standard open file dialog. Users can select any file in their roms folder and the folder is used in the setting. Unfortunately, BizHawk doesn't provide the "select folder" version of the standard open file dialog, so this is how I opted to handle it. Documentation will be updated to help people understand how it works.

Overall, my hope is that this will help even more people get their tracker setup easily. Button configuration will eventually come, but that will be in another PR.


Brian0255:
Looks fantastic! This will be really great for people who don't feel comfortable editing settings files directly, which I think will help out the community a lot.

PR #69, opened 2022-06-04
Reduce updating user data while walking in field

bdjeffyp:
I finally dug into the slowdown issues. I was thinking at first that it was simply a performance issue of individual PCs because it wasn't consistent. Digging more into the code and talking with some users and Besteon, I was able to pinpoint one course of action that should alleviate the issue.

Every step, the tracker was executing the `HandleUpdatePoisonStepCounter()`, triggered by the event of the same name. It turns out that event triggers even if your Pokémon isn't poisoned! So, because this method was called, a tracker redraw event occurs, which in turn makes a call to `Tracker.SaveData()`, which is where all the current tracker data is stringified and stored in BizHawk's `userdata` so that it can be saved in a save state. When the data builds up over time as the player encounters more and more Pokémon, the writes to `userdata` take longer and longer, until the emulator begins to hiccup with those writes.

To mitigate this, I found another memory event for poison: `DoPoisonFieldEffect`. This occurs at less intervals because of the screen distortion effect taking place only every 3-5 steps. Also, I placed a check to see if the lead Pokémon is poisoned. If not, it won't redraw the tracker and write the `userdata`. So, slowdown will now only occur if the player has encountered a lot of Pokémon AND the lead is poisoned when walking in the field.

I'm sure there is more that we can do here, but at least this is a start.

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -80,7 +80,7 @@ function Main.Run()
 		event.onmemoryexecute(Program.HandleCalculateMonStats, GameSettings.CalculateMonStats, "HandleHandleCalculateMonStats")
 		event.onmemoryexecute(Program.HandleDisplayMonLearnedMove, GameSettings.DisplayMonLearnedMove, "HandleDisplayMonLearnedMove")
 		event.onmemoryexecute(Program.HandleSwitchSelectedMons, GameSettings.SwitchSelectedMons, "HandleSwitchSelectedMons")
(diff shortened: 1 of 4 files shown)


valiant-code:
Great find. I wonder, if the performance of writing data continues to become an issue as the tracked data grows, it might be a good approach to make it write to `userdata` only at certain times like at the end of a fight (or another time where even if there is lag, it might not be noticed).

But for now this seems like a good first step to drastically reduce the number of writes.

bdjeffyp:
Thanks! Yeah, reducing the writes will be the next step, and perhaps even chunking the writes in some way, especially if I can grow this out into Gen 4. More Pokémon to track means more data which means even more performance worries...

PR #77, opened 2022-06-12
Pokémon Center Heal Tracker

bdjeffyp:
image
Add full party healing checks to the tracker, showing under the PC Heals value in the tracker.
- Turn this on by selecting the "Survival ruleset" option in the settings menu/Settings.ini file.
- Starts at -1 because the lab gives you a free heal, so successfully escaping the lab will show 0 visits.

Bug fixes:
- Reverted to using the Poison step counter memory stamp while also using the poison status to perform tracker updates and userdata writes
- Check for cancelling file selection when setting the ROMS folder in the settings menu of the tracker

Also updates the readme and increments the release version. I do note that this is not heavily tested yet so there is a chance for problems, so people should be independently tracking until we work out the kinks in this.


GB127:
Shouldn't the playe rbe responsible to keep track of these? Like you add a "+" button to increment wheneve rit's needed and reduce the "listen frequency"

bdjeffyp:
I had considered that. However, we have the potential to see when the player has all of their Pokémon healed so why not take advantage of that to do it automatically? The player's Pokémon page is all autotracked things anyway.

PR #78, opened 2022-06-13
Linux fixes

### LINKED ISSUE #68: Tracker fails to load on first run - Lua Exception
From IronMON Discord: "when loading the .lua... Ironmon-Tracker v0.3.0 NLua.Exceptions.LuaScriptException: A,B,Start,Select"

This has been an ongoing issue for many versions of the tracker, including the latest 3.0.0. Several people each day report this issue. In almost all cases it can be resolved by closing everything and reopening a few times.

I'm assuming the tracker has to initiate some files on first load, but might do so in the wrong order and is unable to find the files when you first run it; hence why closing and opening again fixes it. Is this something that can be resolved, or can a more detailed message be included in the error to instruct the user to restart?

moben:
This might fix #68, at least it does so for me.

After some printf debugging, I noticed that Inifile implementation uses `file:lines()` to read the ini file, but does not break at the `nil` that signifies the end of the file. Instead it unconditionally calls `line:match(…)` [here](https://github.com/besteon/Ironmon-Tracker/blob/2efc900215de2b18d54c3983e87426170f4cdd5a/ironmon_tracker/Inifile.lua#L66), even when `line == nil`.

I have no idea why this only fails sometimes for some people, but for me it was 100%, maybe because under Linux BizHawk uses a different Lua backend.

Also contains a fix to use `/` instead of `\` as the path separator. This fixes loading the next rom under Linux. I believe also works on Windows since 7 or 8, but someone with a Windows machine needs to test this. It did work for me under wine, but that might behave differently in this regard. Otherwise it needs to be some hack like `package.config:sub(1,1)` to be compatible with both.

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -137,10 +137,10 @@ function Main.LoadNext()
 	local romnumber = tonumber(string.match(romname, '[0-9]+')) + 1
 	local nextromname = ""
 	if rombasename == nil then
-		nextromname = Settings.config.ROMS_FOLDER .. "\\" .. romnumber .. ".gba"
+		nextromname = Settings.config.ROMS_FOLDER .. "/" .. romnumber .. ".gba"
 	else
 		rombasename = rombasename:gsub(" ", "_")
-		nextromname = Settings.config.ROMS_FOLDER .. "\\" .. rombasename .. romnumber .. ".gba"
+		nextromname = Settings.config.ROMS_FOLDER .. "/" .. rombasename .. romnumber .. ".gba"
 		print(nextromname)
 	end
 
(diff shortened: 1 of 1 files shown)


moben:
I opted to add the workaround instead of modifying the Iniflile.lua as that seems to be 3rd-party code.

bdjeffyp:
Tested on my Windows 10 machine. It seems good. Thanks for not modifying the IniFile.lua code and implementing the workaround instead.

I did initially get some panic crashes from the Lua engine when trying to load the next seed with the button combination, but after restarting BizHawk and trying again with a controller rather than a keyboard, it worked fine. The keyboard wasn't the issue, because I did try with both joystick and keyboard after restarting BizHawk, so I think it has to do with using the Settings menu to initially set the Roms folder, which I had to do since I was using your code for the first time, which had the default Settings.ini. While the engine panic crash is not associated with the first time run errors brought up in Issue #68, I am happy to see that this will address the issues Linux users were facing.

bdjeffyp (approved):

PR #81, opened 2022-06-14
Improving logic for "Load Next Seed"

UTDZac:
This logic update allows the next seed function to handle ROM names with leading zeros (e.g. "Rom001", "Rom002"). It also handles the weird case of having spaces and/or underscores in the Rom file name, which bizhawk makes difficult. There is a minor amount of error checking done here to report when a "next seed file" doesn't exist, to help with troubleshooting.

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -2,7 +2,7 @@
 -- Created by besteon, based on the PokemonBizhawkLua project by MKDasher
 
 -- The latest version of the tracker. Should be updated with each PR.
-TRACKER_VERSION = "0.3.2"
+TRACKER_VERSION = "0.3.3"
 
 -- A frequently used placeholder when a data field is not applicable
 PLACEHOLDER = "---" -- TODO: Consider moving into a better global constant location? Placed here for now to ensure it is available to all subscripts.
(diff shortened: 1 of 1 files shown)


UTDZac:
@bdjeffyp Tagging you directly to see if you can help with this merge, no worries otherwise.

I wanted to improve on the ability to soft reset and load the next seed in sequence. This is one of the most common issues brought up in the IronMON discord help channel. These fixes should handle a lot of use-cases that people are running into where it breaks the tracker. I tested these myself, here's a example set of seeds I worked with (attached).

image

bdjeffyp:
Dude! This is badass!! I'll review the code tonight after work! Thanks so much!

Ironmon-Tracker.lua
```diff
@@ -132,19 +132,36 @@ function Main.LoadNext()
...
-	client.openrom(nextromname)
+  
+  -- Split the ROM name into its prefix and numerical values
+  local romprefix = string.match(romname, '[^0-9]+')
+  local romnumber = string.match(romname, '[0-9]+')
+  if romprefix == nil then romprefix = "" end
+  
+  -- Increment to the next ROM and determine its full file path
+  local nextromname = string.format(romprefix .. "%0" .. string.len(romnumber) .. "d", romnumber + 1)
+  local nextrompath = Settings.config.ROMS_FOLDER .. "\\" .. nextromname .. ".gba"
```
bdjeffyp, line 143: You may notice that there was a change recently that helped people with Linux by using the `/` rather than `\\`. The `/` works in both Windows and Linux, while the `\\` excludes Linux. Can you swap this line and line 152 to use the `/` instead?

bdjeffyp (changes requested):
Other than the change requested regarding slashes, can you also update line 5 to version `0.3.3`? I intend to get this change out as soon as possible. Also, make sure you load your IDE with the .editorconfig settings. It looks like you are using spaces and Besteon prefers tabs. Thanks!

bdjeffyp (approved):

UTDZac:
Unsure where else to mention this, but the Readme can be updated to remove the requirements for filenames without spaces and with leading zeros.

> Features
> Quick loading seeds: You can create a bunch of seeds ahead of time, and then use a button combination to load the next seed. Seeds must be in a numerical order without leading zeroes.
> For example: you can start at 13 with a file name like kaizo13.gba. Pressing the button combination would then load kaizo14.gba. Press it again and kaizo15.gba is loaded. If you tried kaizo00014.gba, the quick load feature won't work. Remove the leading zeroes.

becomes:

> Features
> Quick loading seeds: You can create a bunch of seeds ahead of time, and then use a button combination to load the next seed. 
> For example: you can start at 13 with a file name like kaizo13.gba. Pressing the button combination would then load kaizo14.gba. Press it again and kaizo15.gba is loaded.

And the FAQ common error can get updated to: 

> Error: NullHawk does not implement memory domains NLua.Exceptions.LuaException: unprotected error in call to Lua API (0)
> 
> Cause: The ROMS_FOLDER path specified in Settings.ini is not correct, or you are out of roms and there is no next seed available.
> 
> Fix: Make sure you have specified a ROMS_FOLDER location. This can be done through the gear icon on the tracker itself.

bdjeffyp:
Sounds good. We can update the README when we next make a pull request. If you want to make the update, go for it. If I try to do it, I have to bug Besteon, who is super busy right now and I won't be able to make a change until a week from now.

PR #83, opened 2022-06-17
Fixing Weight-calc bug and cleaning up ReadMe

UTDZac:
This PR covers three things:

1. There is a minor bug with the weight calculation. It uses less-than comparison instead of less-than-or-equal-to. Calculating base power based on target's weight uses <= according to [The Serebii.net chart](https://www.serebii.net/games/gklk.shtml).
2. Meowth's weight is incorrect. I don't know how this is/was calculated. It should be 4.2kg not 66.6kg. [Source](https://bulbapedia.bulbagarden.net/wiki/Meowth_(Pok%C3%A9mon)).
3. Minor cleanup with the ReadMe instructions and error FAQ. This is from a few updates ago that allows file names to be less strict (no leading zero requirement).

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -2,7 +2,7 @@
 -- Created by besteon, based on the PokemonBizhawkLua project by MKDasher
 
 -- The latest version of the tracker. Should be updated with each PR.
-TRACKER_VERSION = "0.3.4"
+TRACKER_VERSION = "0.3.5"
 
 -- A frequently used placeholder when a data field is not applicable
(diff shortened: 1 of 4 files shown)


UTDZac:
After some discussion with Fellshadow and OnlySpaghettiCode, the util's damage calc formula should be based on Bublapedia's chart (exclusive bounds) and not Serebii's chart (inclusive bounds). This was determined by reviewing the gen 3 decompiled code for Low Kick to see if a weight value of 100 kg should be 80 base power or 100 base power. According to the code itself, its 100.

I've reverted my previous changes to the Utils file in [fcb21aa](https://github.com/besteon/Ironmon-Tracker/pull/83/commits/fcb21aa0f821577f655f37136a71f7f03b646c69) to reflect this, but left the comment explaining what this function does.

PR #84, opened 2022-06-17
Tracker Hotfixes

### LINKED ISSUE #73: Air Lock ability shows up blank
The Air Lock ability does not appear on the tracker, showing up as a blank spot. Since this ability is last in the list in the Data.lua file, my guess is that this is an off-by-one error.

Fellshadow:
fixes #73 

This was in fact an off-by-one error as expected, after looking at the abilities list in the decomp code [here](https://github.com/pret/pokefirered/blob/master/src/data/text/abilities.h) we can see that the ability "Cacophony" occupies the ability slot between shell armor and air lock. 

This ability is an unused soundproof clone meant for the whismur line, bulbapedia seems to have just removed its id in their list of abilities and adjusted air lock onwards accordingly however the game seems to still have a slot/id reserved for it (sBattlerAblities returns 77 for Air Lock but bulbapedia assigns 76 as the id for it). Therefore I went for a simple solution of adding cacophony in the ability data to allow Air Lock to be indexed correctly in the tracker.

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -50,13 +50,12 @@ Main.LoadNextSeed = false
 
 -- Main loop
 function Main.Run()
-	print("Checking for open ROM before loading...")
+	print("Loading...")
 	local romLoaded = false
 	while not romLoaded do
 		if gameinfo.getromname() ~= "Null" then romLoaded = true end
 		emu.frameadvance()
 	end
-	print("Loading...")
 
 	Options.buildTrackerOptionsButtons()
 	GameSettings.initialize()
@@ -126,9 +125,7 @@ function Main.LoadNext()
 
 	if Settings.config.ROMS_FOLDER == nil or Settings.config.ROMS_FOLDER == "" then
 		print("ROMS_FOLDER unspecified. Set this in Settings.ini to automatically switch ROM.")
-		Main.LoadNextSeed = false
-		Main.Run()
-		return
+		Main.CloseROM()
 	end
 
 	local romname = gameinfo.getromname()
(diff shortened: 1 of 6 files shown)


Fellshadow:
Originally this was just a fix for Air Lock but I have now added a couple of other small fixes which are as follows:
-  Added a line that will assign an empty string as the value of `Settings.config.ROMS_FOLDER` if it was read in as nil
    - When the ROMS_FOLDER setting is left empty (i.e. `ROMS_FOLDER=`) the Inifile will read it as a nil value and so `Settings.config.ROMS_FOLDER` does not get assigned a value. 
    - If the user then toggles a setting in the tracker's options menu such that it saves to Settings.ini this results in the ROMS_FOLDER setting getting deleted entirely.
    - This fix allows the options menu to save the empty ROMS_FOLDER setting instead of deleting it.
- Added a check for `Tracker.Data.needCheckSummary` for make the hidden power button visible so it doesn't show before moves are shown. Fixes #86 
- Used an inline if to initialise `Tracker.Data.needCheckSummary` to the correct value depending on the MUST_CHECK_SUMMARY setting
    - When I was testing the hidden power button fix, I noticed sometimes the tracker wouldn't set the flag correctly on loading as it would detect some previous tracker data in memory (perhaps bizhawk auto-loading some old SaveRAM data). 
    - This fix ensures that the needCheckSummary flag initialises correctly

Fellshadow:
Addition: Fixing the check for Bizhawk version

upon testing, I found the string.sub function use to be unnecessary as client.getversion() returns a string of form X.X.X and lua indexing starts at 1, so i've removed this function call

also, significantly older versions of bizhawk do not have a client.getversion() function so included a check for if the function exists

Fellshadow:
New additions:
- Improved error handling on the loading next seed functionality via more helpful output messages for the user.
    - Previously, if it failed to find the next rom the tracker would crash, this change allows the tracker to reload with the current rom, with an output message informing the user.
    - Additionally, a check for if the current ROM has a number in the name has been implemented with a message for the user if they have not loaded a ROM with a number in the name.
- Fixed the tracker crashing if no ROM was loaded in the initial 5s loading wait window
    - The tracker would wait 5s before calling `GameSettings.initialize()`, resulting in a crash and error when attempting to read the ROM memory (as there is no ROM).
    - This fix instead uses `gameinfo.getromname()` to check if a ROM is loaded before it makes the call to `GameSettings.initialize()`, the tracker will currently hang in a "waiting" state until a ROM is loaded

Fellshadow:
Right-justified numbers fix:
- Move power wasn't using the drawNumber function, so wouldn't get right-aligned if the setting was enabled
- The value "---" for moves with no power/accuracy were not being right-aligned correctly

PR #96, opened 2022-07-08
Performance updates

### LINKED ISSUE #1: Own mon ability only updates when it is "used" in battle.
Currently using sBattlerAbilities address to locate mon ability, which is 0 outside of battle, so switching/catching another mon won't update the ability until the mon attacks in battle.

### LINKED ISSUE #3: Track opponent mon ability
When an opponent's ability is activated, reflect that information in the tracker. A mon can potentially have 2 ability slots.

### LINKED ISSUE #13: Trace Ability doesn't track properly
The ability that trace identifies is listed on the tracker as the mon's ability.

### LINKED ISSUE #30: Skill Swap
image

Andy let me know of an issue with Skill Swap. I will look up the VOD of the run in question to see if I can get more details. Logging as a bug separate from the Trace ability issue, but there is a chance it is related to the event handler issue we need the newest BizHawk version to address.

### LINKED ISSUE #42: Ruby/Sapphire support
I been doing over 100  ruby kaizo ironmon seeds its kind of dissapoting that these other gen 3 games are missing unlike emerald, ruby and sapphire have less double battles which makes the games less tedious and I think a tracker update to support these games would be huge

### LINKED ISSUE #49: Tracker refresh of Pokémon midbattle after one is defeated not occurring
During Kelpsey's recent stream: https://clips.twitch.tv/PiercingKitschyPeachFreakinStinkin-9a5EVDmUiUlV3qFa

### LINKED ISSUE #74: Poison update while walking is not occurring
Part of the attempt to reduce the slowdown people have encountered with the tracker involved adjusting the memory stamp that we check for poison damage as the initial value was causing tracker userdata writing to occur with every step.
I adjusted the memory stamp to look at when the screen is distorted. I also added a check for the poison status to be set on the lead Pokémon. If those two conditions are met, then the tracker would redraw, which would also do the userdata write that can result in the slowdown.
Unfortunately, Tso15 discovered that this was not functioning as designed. They were having to open the menu or get into an encounter to see the tracker update. So, I think I would need to restore the original memory stamp that was triggering the event to redraw the tracker while still keeping the check for the poison status. It is also possible that there is a logic issue with the poison status check that wasn't working.
I did test this prior to releasing the change and it seemed to be working properly. I still have a save so I will verify what is happening and see if I missed something.

### LINKED ISSUE #75: Enemy Pokémon showed all moves/stats!
Clip from Tso15 showing this occurring...
https://clips.twitch.tv/SteamyFlirtySandstormTF2John-gGYhi_luFfflOvQL

Great catch finding this. Let's see if we can repro and pinpoint the conditions that caused it. It almost seems like the logic thought that the Sealeo was the player's Pokémon for a moment.

### LINKED ISSUE #87: Transforming into a Pokemon gives you info on their ability that you shouldn't have
This is a bug. Using transform causes some issues with the tracker, one major and one minor, detailed below.

### The major issue
Immediately after using Transform, your Pokemon's ability on the tracker changes to the targeted Pokemon's ability, giving you information you shouldn't be able to access (you can't see it by viewing the Pokemon's summary in-game). Looking at the reference image below, Poliwhirl has the ability Shed Skin, while Charmeleon has Rough Skin.

Is this similar to #30 ?

### The minor issue
After transformation, the tracker continues to show your original Pokemon's moveset and stats. I image not revealing the stats is a good thing. The moveset not reflecting the Pokemon it became is only a small concern; or maybe not one at all. It's not really important to track this data necessarily. However, many people find it useful to know what move is effective against the opposing Pokemon. There may be other reasons to temporarily see the transformed moves show in the tracker.

### Reference image
image

### LINKED ISSUE #94: During Emerald catching tutorial, Wally's pokemon stats are revealed
Able to reproduce in v0.4.1a.

Wally, unlike the old man in FRLG, has a pokemon that he uses to do the catching tutorial. As a result, the tracker seems to think you need to see this information. Probably best to keep this hidden.

image

It's possible some of the reworks to the tracker's data structures I've been working on can have a clean way of resolving this (trainerid maybe?)

UTDZac:
Closes #13, closes #30, closes #42, closes #49, closes #74, closes #75, closes #87, closes #94

This PR more than likely fixes all of the above issues. It only partially fixes #1 as it more reliably updates caught Pokemon abilities. This also fixes #3 for the most part, as it allows for automatically tracking many important ability triggers (some restrictions apply)

This is a **massive rework** of the data structures and logic that the Tracker uses to manage memory reads, tracked information, and much more. It significantly improves performance while attached to Bizhawk. If you have any questions, feel free to ask. I would also encourage any additional testing that you may wish to do.

This update also includes preliminary **Ruby/Sapphire support**. The `Heals in bag` feature is the only one that I know of that doesn't work for these games.

This update now also includes better enemy ability tracking. It will attempt to automatically track some abilities based on the battle message text displayed on screen. Any ability can still be manually tracked by clicking on it's ability area similar to taking notes on a Pokemon.

This update also removes the auto PC heal tracking feature as there is no longer an event trigger. You can still manually track PC heal.


UTDZac:
Some more details on the features that this version update includes can be found here: https://github.com/UTDZac/Ironmon-Tracker/releases/tag/v0.4.2e

bdjeffyp:
I'll check this out and do some testing when I get home from work this evening! I highly encourage getting others involved in testing as well, given the large scope. @Fellshadow, @Brian0255, want to help testing these changes?

UTDZac:
Thanks @bdjeffyp . No rush on getting this out today. I'm still fine tuning some numbers to make things look clean, nothing major.

Fellshadow:
I've already been doing some testing of preliminary builds for this and it was looking great so far (and the issues I found I brought up with zac)
not had a chance to test the 0.4.2e build yet (last one i tested was 0.4.2c) but am down to do more testing when i'm able

UTDZac:
Adding in a new feature that I think many people will really appreciate. Let's you more easily track Pokemon abilities manually, instead of having to use the Notes feature. I *hope* this doesn't break anything, as I only did light testing.

I put a '?' if you only know about one ability, just in case it has a second one. Idk if this question mark is helpful or not, or just has the potential to add confusion.

image
image

UTDZac:
Fixes #94 with [838e5fb](https://github.com/besteon/Ironmon-Tracker/pull/96/commits/838e5fb040fb456a871cc87c8fc581d4738f4587). Thanks Fellshadow!

UTDZac:
Just a quick note here, I'm finally getting around to testing Emerald some more. Many of the delay adjustments I did the abilities (unrelated to auto tracking) seem to not function well in Emerald; different delay numbers. So working on finding a better solution for this.

UTDZac:
This should be fixed now for RSE. There was an annoying bug where `GameSettings.sBattlerAbilities` address value doesn't get updated for a while, using `gBattleMons` instead as it is more accurate. This fixes the issue for the most part, just needs a small delay now (60 frames) instead of a large one (360 frames).

Also added in ability auto-tracking support for Emerald. The other gen3 games (Ruby/Sapphire) look quite doable, just requires a good bit of manual investigation.

UTDZac:
I'm currently in the middle of adding in this new feature. Would still love to get the other bits of this PR reviewed.

### New Feature: Move Info Look-up!
Clicking on a move name on your Pokemon or the enemy Pokemon will open up a new screen that displays information about that move. This uses most of the information that was already on hand as part of the `MoveData` object, but with some additional info added as well.

**New move information added:**
- **Contact**: Is the move a contact move or not.
- **Priority**: Does the move have a priority other than 0, ranging from -6 to +5.
- **Summary**: A short and concise description of relevant info about the move.
  - Currently only gen1 moves have summaries, as they take some time to add in manually. I want to do them manually to ensure only the most useful and accurate data is presented.

The stab text only appears if the Pokemon who has this move matches that move type, otherwise it won't get drawn. Similarly, the line of text showing the `Priority` value won't get drawn for most moves with a default priority of 0.

image

This also adds support for eventually clicking on other things for information, such as the Pokemon icon itself to learn about when it learns all of its moves. Maybe evolutions too.

UTDZac:
The move info look-up feature is completed.

I realize after continually working on this while waiting for the PR review, I probably should have done this new feature as a separate PR. I'll remember that for next time, still getting used to this  whole github thing.

UTDZac:
Unless anything big looks out of order, this release is ready to go and the ReadMe has been updated with the new features and images.

For creating the release, I've listed the big feature updates and additions here, as well as a short version of the install instructions: https://github.com/UTDZac/Ironmon-Tracker/releases/tag/v0.5.0a

bdjeffyp (approved):
Fantastic work! I am excited to see all of this go live!

PR #99, opened 2022-07-14
Fixing visual issue with using special character

UTDZac:
Move data was using "—" to display blank power/accuracy, but should have been using three "-" instead.


UTDZac:
Can merge this PR at any time. These are all extremely minor bug fixes that were caused from me doing tests on Bizhawk 2.9, which handles text rendering and form popups differently, than 2.8

PR #100, opened 2022-07-14
Ruby/Sapphire Summary Check Improvement

Fellshadow:
Previously I implemented a check for the main callback2 function which took a seemingly unique value when in the summary screen of Ruby/Sapphire, however turns out the same value pops up during the initial copyright screen, so the tracker would immediately think the summary was checked

Upon revisiting the [ShowPokemonSummaryScreen function](https://github.com/pret/pokeruby/blob/2162e948d1889cbf5c651b9747dab1194323f2da/src/pokemon_summary_screen.c#L582) in the decomp code, I took a further look at the pssData variable. 

This variable is unique to the summary screen ("pss" = pokemon summary screen) and is stored in gSharedMem [here](https://github.com/pret/pokeruby/blob/a3228d4c86494ee25aff60fc037805ddc1d47d32/include/ewram.h#L130). This got rewritten as sMonSummaryScreen in FRLG/Emerald which functions roughly the same from what I gather

I went with the lastpage value stored in pssData [here](https://github.com/pret/pokeruby/blob/a3228d4c86494ee25aff60fc037805ddc1d47d32/include/pokemon_summary_screen.h#L48) for consistency as it is always set to the contest moves page regardless of where the summary screen is opened from.

This value stays at zero until the summary screen is opened, then takes the value of 3 (index of the contest moves page), so the check for the memory address value can be the exact same as it is for FRLG and Emerald

Current (albeit very minor) issue: 
  - This value remains at 3 permanently, it doesn't get reset, therefore if someone were to toggle the "hide stats until summary shown" setting off and back on, the stats won't be hidden.
  - This also prevents functionality of hiding stats on a new pokemon that has just been swapped to the lead slot until its summary is checked. 
  - To note this issue also happens with sMonSummaryScreen in Emerald however it doesn't seem to be an issue in FRLG

This isn't a very big issue tbf, just noting the behaviour

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,10 @@ This project is based on [MKDasher's PokemonBizhawkLua project](https://github.c
    - **IMPORTANT**: _Run BizHawk once and **close** it_, then start it again before continuing! This ensures that BizHawk sets itself up properly on your system. Otherwise, the tracker may get some odd errors when trying to start it during your first use of BizHawk.
 1. **Download** the project from the [releases](https://github.com/besteon/Ironmon-Tracker/releases/latest) section. The main branch has additional changes and may not be stable.
    - If you are feeling adventurous and wish to help us in finding bugs, you are more than welcome to clone the main branch. If the tracker crashes, please provide the log dump from the Lua Console to us via Discord or the [Issues](https://github.com/besteon/Ironmon-Tracker/issues) tab.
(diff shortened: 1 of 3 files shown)


Fellshadow:
to note, also added a little fix for the README as it refers to the main script file as `Ironmon_Tracker.lua` when the file is named `Ironmon-Tracker.lua` (underscore instead of hyphen)

UTDZac (approved):
This summary check update looks great. Good find on the new memory address that makes this easier.

PR #104, opened 2022-07-15
Relative form location

thisisatest1900:
added function that allows you to set form relative to the top left of the game screen 
second commit uses this function to open the note form at logical for me location and made the note resize to match tracker size when it changes

### DIFF
--- a/ironmon_tracker/Drawing.lua
+++ b/ironmon_tracker/Drawing.lua
@@ -959,3 +959,13 @@ function Drawing.drawImageAsPixels(imageType, x, y, imageShadow)
 		end
 	end
 end
+
+--sets the form location relative to the game window
+--this function does what the built in forms.setlocation function supposed to do
+--currently that function is bugged and should be fixed in 2.9
+function Drawing.setFormLocation(handel,x,y)
+	local ribbonHight = 64 -- so we are below the ribbon menu 
+	local actualLocation = client.transformPoint(x,y)
+	forms.setproperty(Input.noteForm, "Left", client.xpos() + actualLocation['x'] )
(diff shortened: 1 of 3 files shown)


ironmon_tracker/Drawing.lua
```diff
@@ -958,4 +958,14 @@ function Drawing.drawImageAsPixels(imageType, x, y, imageShadow)
...
+end
+
+--sets the form location relative to the game window
+--this function does what the built in forms.setlocation function supposed to do
+--currently that function is bugged and should be fixed in 2.9
+function Drawing.setFormLocation(handel,x,y)
+	local ribbonHight = 64 -- so we are below the ribbon menu 
+	local actualLocation = client.transformPoint(x,y)
+	forms.setproperty(Input.noteForm, "Left", client.xpos() + actualLocation['x'] )
+	forms.setproperty(Input.noteForm, "Top", client.ypos() + actualLocation['y'] + ribbonHight)
```
UTDZac, line 970: Two things here:
1) This function accepts `handel` (typo?) as a parameter but doesn't use it anywhere. I suspect the `Input.noteForm` is supposed to be this parameter, which would allow using this function for other form controls that are rendered within the Tracker. There are a few more I'd like to extend this to.
2) The `Drawing.lua` file ideally handles anything that relates to drawing on the Bizhawk emulation screen, usually in the space reserved for the Tracker. As this function alters something else that is unrelated to drawing, can you move it to `Utils.lua`

UTDZac (changes requested):

UTDZac (approved):

PR #105, opened 2022-07-16
Allowing Tracker data to save to a file

UTDZac:
This is a first step in separating out the tracker data from save-states. There are a couple of reasons for this, as well as potential future improvements on allowing the end-user to better manage tracked data between games.

- The primary reason for separate tracked data is some users do not or can not use save-states. As such, they fear losing their Tracker data and are forced to keep their computer/game running.
- Secondly, some users run into an issue where Bizhawk causes a sound bug when from loading game data from a save state (unrelated to Tracker). They can use in-game saves to bypass this bug, but have no way to get their Tracker data back.

Ultimately this save data feature can be built upon to allow users more control over their data, especially for the purposes of creating backups, for sharing with devs to troubleshoot issues, and for allowing upgrade-safe version updates.


UTDZac:
Finished with this one. Originally wasn't sure if I should bother with giving the user the option to manually save, but after some more thought I built that feature in.

The tracker data "save" behavior is as follows:

- The tracker will auto save its data as a file based on the name of the game (not the ROM name), e.g. "Pokemon FireRed AutoSave.tdat"
- When the tracker is first initialized, it will always attempt to load data from an auto save file, if it exists. Otherwise, it just starts with blank data per normal.
- In addition to the above functionality, the user can manually save data and load data. Saving data defaults to a filename based on the current ROM name, e.g. "FireRed Kaizo5" but can always be edited to any name. Loading data allows for loading any `.tdat` file.

bdjeffyp (approved):

PR #110, opened 2022-07-16
Auto Badge Tracking support

UTDZac:
Thanks to some research from @thisisatest1900 , we can now automatically check for when badges are obtained and auto update the tracker with this information. This feature works for RSE and FRLG.

Also disabling the focus punch check, as that has conditions where it fails and reveals too much info to the user.


Fellshadow (approved):

UTDZac (commented):
Looks good to me.

Issue #37, opened 2022-05-21
Counted heals at top of bag only?


bdjeffyp:
https://clips.twitch.tv/SmilingScrumptiousCookieDoubleRainbow-g7Id7Jdt-ZCR2fg1

Andy got an energy root and it didn't update the heals in bag. Moving the root up the list and exiting the menu resulted in the heal being tracked.

bdjeffyp:
More evidence of the Heals In Bag tracking issues: https://www.twitch.tv/videos/1491289193
I saw the count fluctuating all over the place, even when I wasn't manipulating the position of the items. It does seem like items near the top are more consistently counted, though that specific testing could be checked later to ensure that hypothesis is accurate.

0oF00Do0:
https://clips.twitch.tv/FlirtyDullEndiveHeyGuys-2y5ZmVthnJSIEhl7

I just had something similair happening to me. started right after the battle. before that it was normal. I noticed it later also fluctuating with extreme high numbers, that happened when just walking around random amount of steps.

It seems like there has already been found a solution but somebody mentioned I should post it anyways.

UTDZac:
Watched this bug happen live. It seems during battle the PC Heals were being counted incorrectly (showing 0 total instead of 1 total). 

https://clips.twitch.tv/HonorableFrailOpossumGOWSkull-6J3l4ZiHWkC9IjHq

Oh just now noticed its possible there is a solution for this already. Typed this up already so just going to leave the comment here.

PR #112, opened 2022-07-17
Fix heal count

thisisatest1900:
this is a fix for issue #37  
added function that return the correct saveBlock1Adrress depending if current game is rs or efrlg 
used it to iterate over the bag now
changed badge check to use that function to be consistent

from now on if you need to read something from saveBlock1 use Utils.getSaveBlock1Addr() to get the correct address


UTDZac (commented):
These changes look good and offer some nice cleanup to the saveblock1 stuff.

UTDZac (approved):

PR #114, opened 2022-07-18
Pokemon stadium portraits

Brian0255:
Went ahead and added support Pokémon Stadium portraits created by Amber Cyprian since I think they look super cool. Includes a new options toggle to use them, as well as the folder with all the images in it. I've tested this and there shouldn't be any issues, so please let me know if there are any!


UTDZac (commented):
Requested two minor changes, and have a third larger one:

Once concern I have is that the extension is set when the Tracker starts up instead of being set each time the option is toggled on/off.

Consider the use-case:
- User downloads and loads up new tracker. On load it sets the extension (gif).
- User then downloads the high-quality expansion pack for pokemon stadium images and adds them to their folder. (They do not overwrite any images or remove any old ones)
- **[Issue-1]** These images will never get drawn because the extension was set to gift
- If the user closes the tracker and reopens it, then might expect the new images to load.
- **[Issue-2]** The new png images won't load because the get extension function assumes an order for which is given priority and assumes that order to be GIF > PNG.
   - If anything, the get-extension function should prioritize higher-quality images over the lower-quality ones
   - Ideally, the function could be expanded further to allow the coder to choose (via a parameter?) the ability to specify a priority order (a single option, or a list, either is fine) for which extension to use. Then the function can simply return that extension which can be used by the caller and set however they please.
   - This recommendation might be out of scope for this PR, and thats okay if so. I want to include this at some point in the future.

Am open to discussing this further when I get some more free time later tonight.

UTDZac (approved):
Thanks for the updates. These look good. Glad to get these transparent images added.

PR #119, opened 2022-07-21
Slight cleanups with saveblock accessing

Fellshadow:
The PC Heals update function didn't get included as part of the saveblock1 utils function cleanup so done that
also added a utils function for getting the game's encryption key which is used for both PC heals and heals in bag

(also added a gitignore for the tdat files so we don't need to worry about accidentally adding those in any commits)


UTDZac (approved):
Nice cleanup, and thanks for adding the gitignore

PR #120, opened 2022-07-21
Focus Punch Tracking Fix

### LINKED ISSUE #18: [Move] Focus Punch doesn't track on "warmup"
This may apply to all 2 turn moves. The move is not tracked during the first turn of the move.

Fellshadow:
fixes #18 using the gBattlerAttacker address value (which we already have in game settings)

This value is even if the player is the attacker, and odd if the enemy is. So just added a check so that when the focus punch message pops up only track it if the attacker is the enemy pokemon

I haven't tested this in the case of both player and enemy having focus punch and in the case of doubles but i tested in the lab for myself having focus punch (it correctly didn't track on the enemy mon) and the enemy having focus punch (which it correctly tracked). It should in theory work in doubles too as I essentially check odd/even rather than specifically for a value of 1

Edit: tested in a seed where both me and the opponent have focus punch, me using it didn't incorrectly track on the opponent mon so that's all good


UTDZac (approved):
Good find with the attacker value. That makes so much sense in retrospect.

PR #122, opened 2022-07-22
Refactored Drawing and minor bug fixes

### LINKED ISSUE #121: Loading in a corrupted or malformed TDAT file crashes the Tracker
Version: 0.5.1/0.5.2

The loading of the TDAT file has very little error checking or safety nets. It assumes the TDAT file is written properly (through the pickle process) such that it can then be read (through unpickle). If the user edits the TDAT file, or somehow otherwise corrupts it, then these processes will fail and cause the Tracker to crash. If this was an issue with the AutoSave TDAT file, then it will prevent the user from using the Tracker until they delete the corrupted file.

image
image

UTDZac:
Fixes #121 

Primary focus of this PR is further code organization and cleanup. The main Drawing method that draws the Tracker screen has been split into sections. Move information to be drawn is condensed.

- Moved HiddenPower button stuff off the button and purely into the tracked data.
- Updated multi-hit move summaries
- Added Tracker version to TDAT file
- Fixed issue where loading a corrupted TDAT file crashed tracker repeatedly.
- Also future-proofed loading in mismatched TDAT information: only uses data that fits the current data model.

### DIFF
--- a/ironmon_tracker/Drawing.lua
+++ b/ironmon_tracker/Drawing.lua
@@ -216,11 +216,16 @@ function Drawing.drawPokemonView(pokemon, opposingPokemon)
 		pokemon = defaultPokemon
 	end
 
-	local pokemonInfo
+	-- Add in Pokedex information about the Pokemon
+	local pokedexInfo
 	if pokemon.pokemonID ~= 0 then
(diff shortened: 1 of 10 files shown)


Fellshadow (approved):
Looks good!

PR #123, opened 2022-07-22
Feature: Click on the game screen while learning a new move for tracker info

UTDZac:
Adding a feature that allows the user to click on the game screen itself. If their pokemon is in the process of learning a move, that move will show on the InfoScreen.

Also adding version number to Options screen (finally), and other minor visual updates.

image


Fellshadow (approved):
Nice addition, going to merge but just out of curiosity does this work regardless of where on the screen you click? And is it limited to just the screen you have shown in the image or does it also work when on the summary screen page where you choose what move to delete for the new one?

UTDZac:
Only works if you click on the message bar at the bottom of the game screen, and only when you are in the middle of learning a new move. During the entire move-learning process you should be able to click near the bottom to get move info.

The memory addresses for "LearnMoveLoop" and "MoveReturn" are used to restrict the timing for when these clicks are allowed.

PR #125, opened 2022-07-23
Transform bug fix

Fellshadow:
In the current codebase the transform tracking disable wasn't being correctly reset, meaning moves weren't getting tracked when they should be.

This is because it was only reset when the enemy mon changed, so I've also added a re-set to occur at the end of battle

I also did some cleanup of the transform tracking disable. The second boolean for force-switch moves is no longer required so i removed it. This was initially used because the transform tracking reset used to occur in the "pokeball sent out" event handler which occurs whenever the player or the enemy sends out a pokemon (thus a force-switch would have incorrectly reset the transform tracking disable due to your swapped pokemon coming out)

Tested this out and with this fix it now correctly resets the isTransformed boolean when the enemy sends out another pokemon and at the end of a battle

### DIFF
--- a/ironmon_tracker/Program.lua
+++ b/ironmon_tracker/Program.lua
@@ -10,6 +10,7 @@ Program = {
 	inCatchingTutorial = false,
 	hasCompletedTutorial = false,
 	lastSeenEnemyAbilityId = 0,
+	isTransformed = false,
 }
 
 Program.frames = {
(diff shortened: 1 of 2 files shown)


Fellshadow:
Update: added a small fix for the stat marking buttons state updates. The state itself wasn't being correctly updating between pokemon which lead to weird orders between pokemon after you mark a stat

Example scenario of where it was broken before:
you encounter one pokemon, the button cycles from off -> plus -> minus -> off and you mark a "plus" on the pokemon. You run from / defeat this pokemon and then encounter another pokemon which has no stat markings. The buttons now cycle off -> minus -> off -> plus -> minus instead because it retained the "plus" state from the previous pokemon internally

UTDZac (approved):
Great cleanup on the transform stuff. Thanks

PR #126, opened 2022-07-24
Various 0.5.3 bug fixes and code cleanup

UTDZac:
This fixes some recent bugs introduced with the 0.5.3 update. Many of which resulting from the code refactoring.

Main fixes:
- Text/box shadows are not as harsh on medium/dark backgrounds.
- STAB move wasn't properly being calculated
- Quick-loading on a rom with no numbers now assumes that was rom0, and looks for rom1.
- Settings.ini missing causes a tracker load error
   - Also cleaned out the ini file, moving this info to a Wiki page later on github


Ironmon-Tracker.lua
```diff
@@ -14,56 +19,42 @@ if client.getversion == nil or (client.getversion() ~= "2.8" and client.getversi
...
 	forms.button(form, "Close", function()
 		client.unpause()
 		forms.destroy(form)
 	end, 155, 70)
-	return
 end
 
 -- Main loop
 function Main.Run()
+	Main.LoadSettings()
```
Fellshadow, line 56: This function returns true/false for if the settings are loaded correctly, however this function call here doesn't seem to make use of it? 

Seems like the tracker will just attempt to continue running as normal even if the function here doesn't successfully load the settings (returns false)

Fellshadow (commented):
Great cleanups, just one question regarding the Main.LoadSettings function having an unused return value

Ironmon-Tracker.lua
UTDZac: Ah, this is true. I included the returns as an optionally way to use `LoadSettings`, in case we wanted that in the future. THey aren't needed now, as there is enough error checking and safety nets to prevent anything bad from happening during the loading of the settings. As such, I can simply remove the `return`s, but also don't think they're hurting anything by staying in?

Fellshadow (approved):

Issue #62, opened 2022-06-01
Support for non-English Pokémon versions


Siphonay:
Hi! I intend to strat streaming IronMon to my Fracophone audience, and as such I would like to be using the French version of the games. It would be even better if the tracker had translated Pokémon, abilities names…

I tried to edit the code to force the script to load with a French LeafGreen ROM, trying 1.0 and 1.1 variables, unfortunately the tracker does not work beyond detecting the Pokémon species and crashes the emulator in battle, so I assume memory addresses are different for every release.

Besides English, Pokémon FR, LG and E have been released in the following languages:
- Japanese
- French
- German
- Spanish
- Italian

I looked around at the code and I think (this maybe wrong or incomplete) implementing multi-language support would entail f all required memory addresses for all revisions of every language release (according to TCRF, the Japanese version has V1.0 and V1.1 releases, while the European languages did not receive a revision)

If we also want to go the route of adding names of Pokémon, Abilities, Moves, Items, Natures, Statuses and types in several languages (note that Spanish and Italian use English Pokémon names), they would need to be added to the data files.

I would definitely be willing to contribute to this should you decide that this is within the scope’s project. Depending on the approach for doing the second part, I could try to write automated scripts with PokéAPI to add variables with names in other languages.

If you know how to find the memory addresses, adding them to the GameSettings.lua would be trivial if not a little time-consuming, but unfortunately at the moment I wouldn’t know where to look for to find the right addresses, but this is something I can learn.

Sorry, this is very wordy! As this is something I really want to see added to the tracker, I made sure to research the feasibility, and as I said I would be willing to contribute as much as possible to realize it. I think it would help the IronMon open up to broader audiences outside of the Anglophone streamers and players. I would understand if you would consider this too heavy to support though − in that case I would certainly do my own fork that supports at least French language games.

Thanks for reading!

GB127:
I believe the best would be to not use an api while running this. Some users have noticed some lags after logging a couple of notes. So we want to reduce the CPU power to a minimum while running the game.

bdjeffyp:
Oh yes, we absolutely will not use APIs to fetch data. All data must be stored within our *Data script files.

Siphonay:
Oh yeah definitely @GB127, sorry I didn't make this clear enough, I wanted to mean that I'd use an API in a separate one-shot script I'd write to add the multilingual data to the files instead of doing it manually, but that wouldn't be shipped with the tracker

[...]

PR #127, opened 2022-07-25
Spanish and french support

thisisatest1900:
i have added the memory values for both spanish and french fire red games 
this include the ability memory values 
the abilities offsets are different from firered1.1 to 1.0 so i expect some of them are wrong
i tested a few but not all of them

i personally tested both games to a bit after brock and didn't see anything wrong 

the main difference is the gSaveBlock1ptr and gSaveBlock2ptr and looks like they are the same for spanish and french 
and the abilities offsets there it seems that the offset are different between games but between each ability it looks like the same difference for example between speedboost to drizzle there is 27 difference in eng fire red and same difference is in spanish fire red

### DIFF
--- a/ironmon_tracker/GameSettings.lua
+++ b/ironmon_tracker/GameSettings.lua
@@ -106,7 +106,7 @@ function GameSettings.setGameInfo(gamecode)
 		},
 		[0x42505253] = {
 			GAME_NUMBER = 3,
-			GAME_NAME = "Pokemon Rojo Fuego (fr spanish)",
+			GAME_NAME = "Pokemon Rojo Fuego (Spain)",
 			VERSION_GROUP = 2,
 			PSTATS = 0x2024284,
 			ESTATS = 0x202402C,
@@ -115,7 +115,7 @@ function GameSettings.setGameInfo(gamecode)
 		},
 		[0x42505246] = {
 			GAME_NUMBER = 3,
-			GAME_NAME = "Pokemon Rouge Feu (fr french)",
+			GAME_NAME = "Pokemon Rouge Feu (France)",
 			VERSION_GROUP = 2,
 			PSTATS = 0x2024284,
 			ESTATS = 0x202402C,
@@ -503,7 +503,8 @@ end
 
 function GameSettings.setGameAsFireRedSpanish(gameversion)
 	if gameversion == 0x005A0000 then
-		print("ROM Detected: Pokemon Rojo Fuego (fr spanish)")
+		print("ROM Detected: Pokemon Rojo Fuego")
+		GameSettings.gBaseStats = 0x082547f4
 		GameSettings.sMonSummaryScreen = 0x0203b140
 		GameSettings.sSpecialFlags = 0x020370e0 -- not sure if its the real value used for.its used for rse only anyway so not that important
 		GameSettings.sBattlerAbilities = 0x02039a30 --not used in tracker so no idea
@@ -581,7 +582,8 @@ end
 
 function GameSettings.setGameAsFireRedFrench(gameversion)
 	if gameversion == 0x00670000 then
-		print("ROM Detected: Pokemon Rojo Fuego (fr french)")
+		print("ROM Detected: Pokemon Rouge Feu")
+		GameSettings.gBaseStats = 0x082547f4
 		GameSettings.sMonSummaryScreen = 0x0203b140
 		GameSettings.sSpecialFlags = 0x020370e0 -- not sure if its the real value used for.its used for rse only anyway so not that important
 		GameSettings.sBattlerAbilities = 0x02039a30 --not used in tracker so no idea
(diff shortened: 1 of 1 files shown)


UTDZac (commented):
These look great. Two thoughts:

1) Do you happen to have the resource for where you got these values for? I have the ones for the English FRLG games, but didn't know the memory address information was available for other languages. Would be interested in seeing the source of these two games, and any others you might have on hand.
1a) As a followup sometime later, I want to get `GameSettings.gBaseStats` included in these two game additions. See #129 

2) Given the names used for these games, it currently breaks my very simple AutoSave name generator (at the bottom of GameSettings). Ideally this would just strip out the part in parenthesis. I can handle this later no problem, will just require some testing on my end.

Fellshadow:
Agree on Zac's thoughts there, also to add: Ideally long-term perhaps we could look into some means of having the tracker itself match the language of the game? (not a task to undertake right now, just thinking ahead. Special characters are of course the big issue there)

Also with the error message popups I added an error message popup in `GameSettings.lua` for when the game isn't a match for what we have supported rn. It currently sends the string `"This game is unsupported.\n\nOnly RSE/FRLG English versions are currently supported."` to the `Main.DisplayError` function, should perhaps change that message since not only English versions would be supported with this addition lol

maybe can add currently supported versions to the `Readme.txt` that comes with the tracker download (and also the main `Readme.md`) and change the string to something like `"This game is unsupported.\n\nCheck the Readme.txt file for supported versions."`?

Mentioning the non-English game support issue here so anyone looking at that issue can see this PR and the progress on the issue - #62

thisisatest1900:
for the rom names i gave them the name i though would help with debugging 
the actual rom names are 
Pokemon - Version Rouge Feu (France)
Pokemon - Edicion Rojo Fuego (Spain)
i can change the name to this and i could update the AutoSave function to just delete the (lang) if you want to

for the error msg we could add RSE/FRLG English versions or fr spanish and french if you don't want to deal with extra files
though adding other languages from now shouldn't take that long so it could grow if i have the time and ppl would want it 

how i got the values for the addresses 
i had to figure the values out myself i didn't find anything on my brief internet search
the tldr is other languages games are based on fire red 1.1 so i expected a lot of the value to be close to those
so i expect that gBaseStats would be the same as firered1.1 so adding it shouldn't be a problem

i will explain how i found out the values here this is a bit wordy and most likely wont be needed for further support so sorry in advance 

to figure out the values i used 3 different technique

the first one for addresses that are static and dont change over time or after game reset and i can control the values of for example sMonSummaryScreen
 or gBattlerPartyIndexesSelfSlotOne
i used the ram search tool in bizhawk 
for example gBattlerPartyIndexesSelfSlotOne changes to 1 if you use the second pokemon in your party and 0 if its the first one 
after doing it for few values i noticed that all of those values were the same as fire red 1.1
so values like that are expected to be the same

second technique for ptr values addresses that change where they are pointing over time or after game reset eg gSaveBlock1ptr gSaveBlock2ptr
i used this technique  
for gSaveBlock1ptr the first 2 16bit values are the x and y coords on the map so i went to corner of the map in regular fire red and checked those value 

then i went to the same corner on the spanish fire red and bizhawk has option in hex editor to save bin file of the memory region so i saved the ewram and iwram 
i know that  gSaveBlock1 is saved in ewram so i searched in the bin file for those coords 
when you find it you look at the address that it found at for example 2552c
i know that actual gSaveBlock1ptr is saved in iwram (because it has 3 in highest byte in eng fr) and pointers are saved in little endian so you search for 2c 55 02 in the iwram file the address that this is in the iwram file is the adress you want for gSaveBlock1ptr just need to add 3 to the highest bit so if i found 4f58 the address to search is 0x03004f58

same method i used for finding gSaveBlock2ptr but used the trainer id for the inital value search instead of coords 
after doing that i noticed that gSaveBlock2ptr is just +4 from gSaveBlock1ptr so no need to work hard for it

third technique to find the abilities those were the trickiest so i expect to have few wrong ones there 
before this code:
```
if opposingPokemon.abilityId ~= ownersPokemon.abilityId then
	-- Only track the triggered ability if that ability belongs to the enemy Pokemon (matches its real ability)
	if GameSettings.ABILITIES[battleMsg] == opposingPokemon.abilityId then
		Tracker.TrackAbility(opposingPokemon.pokemonID, opposingPokemon.abilityId)
	end
end
```

i added 
```
if(battleMsg ~= battle) then
	print("pointer: " .. string.format("%x",battleMsg))
	battle = battleMsg
end
```
this showed me in every battle all the battleMsg so i tested drizzel and for example got 3 possible values
then i expected that speedboost would be on the same distance as it is from the firered 1.1 version (the distance is 27)
so i generated a seed where i had kyoger(drizzel) and ninjask (speed boost) as starters and searched for 2 battlemsgs that the distance between them were 27 i did it for damp and flash fire aswell to be sure 
i expect that in the abilities there would be some difference because for firered 1.0 there is some difference  but from the abilities i checked i didn't see any 

this is long but i think could be of help for other ppl that want to further support and just document it so i wont forget myself as well

thisisatest1900:
i did a bit of digging around why the é doesnt render correctly 
from my understanding lua strings are array of char and char is 1 byte value and this is what passed for gui.drawtext so we could only use utf8 chars to draw in that way so for example we cant draw Japanese characters with that method

for some reason if you use "Poké Doll" as static string it translate the é to the byte value of c3 
é in utf 16 which is what c# string are should be 0xe9 
so if i use `"Pok" .. string.char(0xe9) .. " Doll"` it does work 
image

Fellshadow:
> for the error msg we could add RSE/FRLG English versions or fr spanish and french if you don't want to deal with extra files
though adding other languages from now shouldn't take that long so it could grow if i have the time and ppl would want it

The issue with this approach is that there's already not much more room in the error popup for the extra message length, and I think long-term would be better off the way I proposed since there's already a Readme.txt file that gets included in the download. Don't worry about implementing this now though, I can do it somepoint soon was more just noting it as a thing lol

Nice finds though with the memory addresses and the é character, I imagine this'd also extend to other special characters too?

UTDZac (approved):
These updates look great, thanks!

PR #129, opened 2022-07-25
Outside of Battle Ability Lookup (finally)

UTDZac:
By looking up ability info from the base stats memory area, we can determine the randomized ability outside of battle. Only catch is figuring out if its the 1st or 2nd ability, which is info we already have.

What this effectively will allow us to do is show the Pokemon's ability on the tracker as soon as the user selects their starter, without having to wait to enter battle. The same is true for any gift Pokemon. Therefore, ability information should be more accurate.

This should work for all games, FRLG and RSE.

image

### DIFF
--- a/ironmon_tracker/GameSettings.lua
+++ b/ironmon_tracker/GameSettings.lua
@@ -7,6 +7,7 @@ GameSettings = {
 	pstats = 0,
 	estats = 0,
 
+	gBaseStats = 0x00000000,
 	sMonSummaryScreen = 0x00000000,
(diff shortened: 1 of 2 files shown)


Fellshadow (approved):
Great find, gonna hold off on merging this in until #126 gets merged since line 214 of Program.lua (`local miscoffset   = (MiscData.TableData.misc[aux + 1] - 1) * 12`) gets commented out in that PR (don't want that commented out in the final main branch lol), and there's a question I have on that one before confirming it for merge.

Other than that I don't think there'd be any issues merging this after that one, when looking between the two PRs

PR #131, opened 2022-07-26
Bugfixes and Friendship Evo update, preparing for 0.5.4 release

### LINKED ISSUE #128: Lowered friendship requirement from the new randomizer setting is not reflected on Tracker
**Version**: v0.5.3

The Tracker currently informs the user if their lead Pokemon is ready to evolve. This is especially helpful for friendship evolutions where the Tracker knows the Pokemon's current friendship value. In the base game, different Pokemon may have different starting base friendship values (normally 70), but they all evolve at friendship value of 220.

As of Randomizer ZX 4.5.1 there exists a setting to speed up friendship evolutions. Unfortunately for us, this setting changes the value required for evolving (160 instead of 220) instead of the base friendship value when caught. Therefore, if the user is playing on a ROM with this adjusted setting and they get their Pokemon to the required friendship value (160), the Tracker still won't show it as ready to evolve, since it looks for the original value of 220.

Source: https://github.com/Ajarmar/universal-pokemon-randomizer-zx/issues/485#issuecomment-1185646762

I'm not sure where this required friendship value for evolving would be, and expect it to be different for other friendship related things like Return and Frustration.

UTDZac:
Resolves #128 

This PR covers a variety of minor bug fixes and a small update to the friendship evolution check to accommodate a randomizer setting.

- ColorPicker distance function bug caused by previous cleanup
- Visual updates to Move Info screen
- Renaming a theme option to be easier to understand (requires awkward workaround)
- Cleaning out some comments
- TDAT files now share data on multiple lines (vs single line)

### DIFF
--- a/ironmon_tracker/Utils.lua
+++ b/ironmon_tracker/Utils.lua
@@ -184,7 +184,11 @@ function Utils.getDetailedEvolutionsInfo(evoMethod)
 	end
 
 	if evoMethod == PokemonData.Evolutions.FRIEND then
-		return { "220 Friendship" }
+		if Program.friendshipRequired ~= nil and Program.friendshipRequired > 1 then
+			return { Program.friendshipRequired .. " Friendship" }
+		else
+			return { "220 Friendship" }
+		end
 	elseif evoMethod == PokemonData.Evolutions.STONES then
 		return { "5 Diff. Stones" }
 	elseif evoMethod == PokemonData.Evolutions.THUNDER then
--- a/ironmon_tracker/Utils.lua
+++ b/ironmon_tracker/Utils.lua
@@ -407,9 +407,11 @@ function Utils.writeTableToFile(table, filename)
 	if file ~= nil then
 		local dataString = Pickle.pickle(table)
 
-		if dataString:sub(-1) ~= "\n" then dataString = dataString .. "\n" end --append a trailing \n if one is absent
+		--append a trailing \n if one is absent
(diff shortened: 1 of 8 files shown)


ironmon_tracker/Utils.lua
```diff
@@ -184,7 +184,11 @@ function Utils.getDetailedEvolutionsInfo(evoMethod)
 	end
 
 	if evoMethod == PokemonData.Evolutions.FRIEND then
-		return { "220 Friendship" }
+		if Program.friendshipRequired ~= nil and Program.friendshipRequired > 1 then
+			return { Program.friendshipRequired .. " Friendship" }
+		else
+			return { "220 Friendship" }
+		end
```
Fellshadow, line 191: This else block here wouldn't ever activate since `Program.friendshipRequired` only gets changed in `Program.Initialize` if the read value is between 1 and 220, and is initialised to 220 in the definition of the Program variables

Not an issue, and still a potential safety net if something goes horribly wrong i suppose, but just noting that this if else is made redundant by previous safety checks

Fellshadow (approved):
Looks good!

ironmon_tracker/Utils.lua
UTDZac: Aye, thanks for pointing this out. I did realize this as I was writing it, but decided to leave it in just as a Default value in case someone else changed something, or something weird happened with setting the friendship value.

PR #133, opened 2022-07-27
fix ability blanking out on evolution

Fellshadow:
Leftover code from the old battle-reliant ability reads was what caused the abilities to blank out and break in the new ability-reading system.

It'd set the abilityId to nil on evolution, meaning the later part updating the pokemon data with the data from the memory read doesn't update abilityId because pokemon[abilityId] == nil. 

This was fine when abilities were read separately in battle as it ran a nil check but now abilities are read along with the pokemonData from `Program.readNewPokemonFromMemory` so it getting set to nil when the pokemonId changes caused abilities to be skipped over when updating data from the newly read pokemonData

(note: I've already slid this into the current v0.5.4 release)

### DIFF
--- a/ironmon_tracker/Tracker.lua
+++ b/ironmon_tracker/Tracker.lua
@@ -26,12 +26,6 @@ function Tracker.addUpdatePokemon(pokemonData, personality, isOwn)
 
 	-- If the Pokemon already exists, update the parts of it that you can; otherwise add it.
 	if pokemon ~= nil then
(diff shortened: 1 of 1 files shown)


UTDZac (approved):
Agree this should fix this issue. I have a follow-up fix for when the ability initially gets read from memory to verify its not blank (checks the other abilitynum option), which i'll include later. But your fix should cover most cases.

PR #136, opened 2022-07-31
Localization france spain

thisisatest1900:
this adds pokemon,move,ability,items as they are displayed to the player in France and Spain versions

the way i found to bypass the wrong use of non standard ascii chars in lua is by writing the byte value of the char so for example:

"Doblebofet\243n", --Doblebofetón english:DoubleSlap 
(look at the ó in utf-16 it has the value 243 c# string are utf-16)

to make it more readable to us and easier to search for everything 
i followed the format above declaring the string with the \### to write the non ascii char with comment of how it looks and its translated form in english
in the france version its less noticeable because they only use standard ascii chars but in the spain version it makes it more readable

the first commit was to add missing item oak parcel without it it messed with how i add items and fixed the é for both poké ball and doll 

i have not included the script that i used to scrape the data because i don't think it should be added to tracker (its one time use thing)
but i can add it if you want to


UTDZac (commented):
This looks great and I'm happy to get this included. A few things:

- I think the Pokemon names for the SpainData file are missing? Unsure if this is intentional.
- Can you create a subfolder and place these two language data files inside. This helps keep the root level structure of the project easily accessible. Something like
   - `/ironmon_tracker/languages/SpainData.lua`

thisisatest1900:
- they didn't translated pokemon names for spain version so they are 1:1 english version so there is no reason to change it
- good catch missed it will fix it and missed Pokéblock Case aswell
- yea that not a problem

UTDZac (approved):

Fellshadow:
looks great and clean, mentioning #62 here to note progress on non-english support

PR #137, opened 2022-08-01
Update currently supported games

Fellshadow:
Added currently supported games to the README files, and updated the unsupported game error message to point the user to the README.txt file for supported games

Also did a little bit of cleanup in general on the main github README, the last error listed on the FAQ is outdated since it related to the quick-load functionality which now has its own error handling to cover this.


UTDZac (approved):
Good call on getting these updated with all the new supported versions getting added.

PR #139, opened 2022-08-02
Add Updated (Gen 7+) portraits option

kittenchilly:
These sprites come from [https://msikma.github.io/pokesprite/index.html](https://msikma.github.io/pokesprite/index.html) which are in turn taken from the generation 7 and generation 8 games. Is an option in the same vain as the Pokemon Stadium portraits option.


UTDZac:
Thank you for doing this. I like the idea of having even more portrait options available.

Going to wait on this one for a few reasons:
- The Options screen has gotten squished will all the extra options. I was planning on building out the menu a bit better.
- I have a fairly large Carousel feature that will be added soon and will also need to have these icon / path updates applied to it.

UTDZac (approved):
Thanks for getting all of these images made, looks like a ton of work. Glad to have these included!

PR #140, opened 2022-08-02
Italy support

thisisatest1900:
this adds italy support for fire red #62 

the abilities section i am guessing that small portion of them are wrong(wouldn't track it but would track it in english version) but didn't want to take too deep dive into it before the new ability tracking logic is merged 

i fixed few wrong addresses in the other languages 

my understanding what addresses are different between english to other languages so far so it would be easier to update:

- addresses that in ewram (have 0x02 as high bit) should be the same between fr 1.1 to all other languages

- addresses that in iwram (have 0x03 as high bit) are different between fr 1.1 to other langueges but same between them 
   so for example gsaveBlock1ptr has same address between spain italy and france

- addresses that in rom (have 0x08 as high bit)  are different between everything but most of them you can search for same sequence of bytes between roms to find them so you go to the address in fr 1.1 look at 10 bytes there and search for them in the other language can use the hex editor in bizhawk or something else hex editor in bizhawk is a bit buggy though this worked for few of the abilities but not all of them


Fellshadow (commented):
nice work! Glad to get this added in, and thanks for the explanation about addresses too (seems like transferring over the new abilities stuff is going to be a pain lol but we'll deal with that when we get there)

Just one small nitpick before i merge this in, where you added the game to the README files could you perhaps add the é in "Pokémon" and remove the hyphen in the game name, just so it's formatted the same as the other games in the list?

Also, would you expect things to differ much between FR and LG for the different languages? May be nice to add in support for those too somepoint, though i'm purely asking as a curiosity right now so don't feel like you need to work on those immediately

Fellshadow (approved):

thisisatest1900:
about abilities i have few ideas how to do it and most of them should work as is so when your branch gets merged i will look into it 
for lg i have not looked into it at all but i expect it would be the same main hurdle would be finding the abilities other things i expect to be pretty easy 
i made few scripts to help with that but they are mainly scripts that only i would understand cause i just made them just to get it done

Fellshadow:
ok doke fair enough, in which case what i'll do when updating the abilities lists for the different games is just update the non-english versions to the same structure as the other games so that it at least doesn't break, and leave comments for what's missing and can fill in the gaps later

PR #142, opened 2022-08-03
Status in tracker

jtigues:
- Active pokemon's Status is now drawn over the top right corner of the pokemon's icon while out of battle
- Added comments on the status logging
- Adjusted the Status enum and added Status-to-Status-Code map


UTDZac:
Thanks for getting this added. A simple addition that goes a long way.

PR #143, opened 2022-08-04
Auto-ability tracking

Fellshadow:
Adds a lot of logic checks to determine it was the enemy's ability that went off when the relevant battle message occurs, to reveal the enemy's ability on the tracker.

There's a chance that there's some weird edge-cases we missed here, can iron those out if found, and can definitely be cleaned up / refactored later but this is seemingly in a functional state. Only FireRed 1.0 and 1.1 have been extensively tested but from some very light testing seems functional on other games (at least it doesn't seem to crash anyways lol)

big thanks to @jtigues for their help with covering as many edge-cases as we can, and on the FR 1.1 side of things. Also thanks to @thisisatest1900 for help with the non-english FireRed versions


UTDZac (approved):
Awesome work. Lots of new code here. Thank you and everyone else who worked on testing this. Looks great.

PR #144, opened 2022-08-04
fix egg spoilers

Fellshadow:
Currently there's an issue where if you put an egg in the lead spot of the party the tracker reveals exactly what mon it is, as though it wasn't an egg
Screenshot_13
Luckily the pokemon data we read also includes a bit for if they are an egg or not, so this issue is fixed by simply cycling to the next non-egg party member if the one that's currently selected is an egg (i.e. the tracker will show the pokemon that the game would send into battle if you entered one). You are always required to have one non-egg pokemon in the party.
Screenshot_14
Screenshot_15


UTDZac (approved):
Good find with the egg reveal. Straightforward fix.

PR #145, opened 2022-08-04
Carousel Feature, Full Route Data, and Last Damage Tracking

UTDZac:
This is a fairly large feature addition that includes:

- A rotating carousel that occupies the bottom box on the main Tracker screen
- All encounter data for wild pokemon in every route for FRLG and RSE (it's a lot of data)
- New Route Info screen to see what wild pokemon have been encountered thus far
- Last damage taken from an enemy move during battle is shown to the player in the carousel
- Helpful tips (could use more additions to this) are shown at the start of each new game
- Status labels on top of Pokemon icons. Thanks @jtigues
- Settings.ini file is no longer included in the project/release. It will still be auto-generated if any settings change.
   - This allows for a more seamless upgrade process for the end-user, allowing them to simply overwrite all old files

Slightly outdated image preview here, but still mostly accurate
image


Fellshadow (approved):
Looks great, can't wait to see people using this!

PR #146, opened 2022-08-05
New Navigation Menu and bonus features

UTDZac:
This adds a brand new navigation menu for getting around the Tracker's Settings. This should give extra space for adding new options in the future. I also clarified how some other options work, like how the Tracker actually saves data.

This should finally allow for #139 to get added in, after some minor conflict cleanup. The location of most icon paths have changed, but it's much easier now to add in new Pokemon icon sets.

Oh, and finally got around to adding in a Credits screen.

image


Fellshadow (approved):
Looks good, wonder when someone will first notice the mirage button 👀

PR #147, opened 2022-08-05
Automatic Weather Ball Updates

### LINKED ISSUE #113: Weather Ball ideally automatically changes type and power in the weather
This is more of an investigation issue to see if it's at all possible to do.

The move Weather Ball changes type and power in the weather. Notably, if the weather isn't normal (or fog), it changes to 100 power and a type that matches that weather. Currently the user can learn about this information simply by clicking the move on the screen, however it would be convenient if the weather information could be read from the game and reflect in the weather ball move itself.

This is likely a two-part issue:
1. Determine if its possible to read in weather information during battle, and remain accurate even if the weather changes.
2. Rework how move data is displayed/drawn to the tracker screen. By default it matches a move id with the static move data, but we ultimately want this to be dynamic. I'm imaging a rework on how the move data is first calculated, then later the Drawing operation can simply draw what has already been calculated. This part is a good chunk of work.

EDIT: Adding in a note this "feature" should be part of the calculate variable damage option, so players who dont want the extra help can opt-out.

Fellshadow:
Resolves #113 
- Adds function that automatically adjusts weather ball's type and power for the current weather in-battle.
- Also updates it in the move info page.
- Weather ball stays at vanilla type/power if not in battle or if the "Calculate variable damage" option is disabled

Screenshot_17
image

Only current issue relating to this is that if Castform has its vanilla ability Forecast then STAB still gets calculated based off of castform's normal form:
Screenshot_16

Given that forecast *only* works on Castform and I believe it's removed from the randomisation pool for this reason, this specifically is a lower-priority issue though would be nice to fix for casual playthroughs.

This is an overall Pokémon type-tracking issue though since the same would happen in the case of Color Change or moves like Conversion.

### DIFF
--- a/ironmon_tracker/GameSettings.lua
+++ b/ironmon_tracker/GameSettings.lua
@@ -29,6 +29,7 @@ GameSettings = {
 	gMoveToLearn = 0x00000000,
 	gBattleOutcome = 0x00000000, -- [0 = In battle, 1 = Won the match, 2 = Lost the match, 4 = Fled, 7 = Caught]
 	gMoveResultFlags = 0x00000000,
+	gBattleWeather = 0x00000000,
 
 	gMapHeader = 0x00000000,
 	gBattleTerrain = 0x00000000,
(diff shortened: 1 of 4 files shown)


UTDZac (approved):
Awesome inclusion, thanks for doing this

PR #148, opened 2022-08-05
Last damage warning

Fellshadow:
This was a suggestion given in the discord, to change the text colour of the last damage taken carousel text to warn the player that the last damage taken is potentially lethal
Went with negative text colour for this, but if another colour is preferred then lemme know
image


UTDZac (approved):
Looks like a fun little add-on to the Last Damage feature. Good idea.

PR #150, opened 2022-08-05
German support

thisisatest1900:
german support for fire red #62

this is probably the last language we could add to fire red only language that is left is japanese and we cant draw those characters on screen currently so i don't know if anyone would be interested in that 

last commit was the manual merge resolve i hope i didn't mess anything there


UTDZac (approved):
Awesome once again. Thank you so much for these additional languages. I'm going to sneak this in to the 0.6.0 release.

PR #152, opened 2022-08-06
DoubleBattleBug

jtigues:
- A number % 2 is never 3, I think this might have been some copy pasta


UTDZac (approved):
Good eye to catch this one.

PR #153, opened 2022-08-06
Fixing infinite loop crash and blinking carousel

UTDZac:
There is an issue with the isEgg memory read where if the wrong value is read from memory (caused by game data hiccups), then it will cause the getPokemon to get stuck in an infinite loop and crash Bizhawk.

This fix ensures the loop eventually breaks by looping back around to the original viewslot.

I've also included a fix for users who regularly use speed-up and notice the carousel rotation speed causes it to blink.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ This project is based on [MKDasher's PokemonBizhawkLua project](https://github.c
 
 ## Supported Games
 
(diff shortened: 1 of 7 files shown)


UTDZac:
I've already updated the 0.6.0 release download with this priority fix.

Fellshadow (approved):

Issue #130, opened 2022-07-25
Tracker doesn't accurately reflect move info that has been randomized


UTDZac:
**Version**: v0.5.3

The Randomizer ZX program has the ability to randomize the properties of a move. These are:
- Power
- Accuracy
- PP
- Type

The Tracker simply uses the vanilla game move data, and as such won't accurately reflect the move data for games where move info has been randomized.

I believe this can be accomplished by reading in the data from the game memory itself, similar to how we now read in ability data for each Pokemon, using the information found here: https://github.com/pret/pokefirered/blob/52c25bac8e7c9fe7ad97d7890be89c7feededf65/src/data/battle_moves.h

The only trick would be if we wanted a way to check first if this data has been altered before we perform all of the memory reads necessary to load in this move data; unnecessary in the common case of move info not being randomized.

Fellshadow:
side-thought about a possible extra use-case for this: if reading move data directly from the game instead of hard-coded could this also be used for multi-language support? 

The idea being that if we're reading move names from the game directly then it'd automatically be in the same language as the game and therefore displayed in the right language on the tracker (though would need to set up the drawing functions to handle special characters in these)

thisisatest1900:
i played with it a bit battlemove struct is declared as 9 bytes but in actuality it takes 12 bytes last 3 bytes seems to always be 00 so when iterating over the array should skip those bytes 
my though if there is no batter way you can read like 3-4 moves and check if all the power pp acc and such are the same it would be very unlikely that even if you randomize only one of those things all the start moves would have same stats

about the names it could be good idea this is where the movenames are stored
https://github.com/pret/pokefirered/blob/master/src/data/text/move_names.h
i think this is one of the address that is different between rom so would need to find a way to find it in rom i couldnt find it in the spain version with intuitive search 

the way that the game store characters is not the same as ascii or utf8 so it would need to pass through some kind of translation for example A is BB it looks like B is BC so for english alphabet its not a problem 
but for É i think is 08 (i am not sure cause i havent found move names yet) so probably would need to find all the special chars

PR #154, opened 2022-08-07
Support for randomized Typings

UTDZac:
This idea came from a user on the Ironmon Discord.
> If the Tracker is able to show accurate typings for Pokemon that have had their typings randomized, then this allows for a fun new gameplay experience. Normally it's frustrating playing against random typings, as everything is a black-box and you're just "Hidden Power" guessing. If you know the types up front, then it plays like a normal Pokemon game where you can make strategic decisions about what moves to use. In other words, this adds a fresh gameplay experience of getting to play with your favorite Pokemon but with better, or worse, typings.

This PR adds support for the Tracker to handle changed Pokemon Pokedex data, such as core Pokemon having different types (and abilities). I've included the abilities here as well so that we don't have to read them in every 30 frames.

Ideally in the future this can be extended to read in MoveData as well, but have to be careful not to overwrite the existing localization data.

### DIFF
--- a/ironmon_tracker/Constants.lua
+++ b/ironmon_tracker/Constants.lua
@@ -1,6 +1,7 @@
 Constants = {}
 
 Constants.BLANKLINE = "---"
+Constants.HIDDEN_INFO = "?"
 Constants.NO_POWER = Constants.BLANKLINE
 Constants.NO_PP = ""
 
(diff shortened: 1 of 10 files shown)


UTDZac:
Likely related to #130 as a building-block for getting that added.

UTDZac:
As requested, adding an Option to enable revealing the randomized data to the player (Pokemon Types and Move Info).

image

Fellshadow (approved):
Looks great!

PR #155, opened 2022-08-07
Make Flail calculation more accurate

kittenchilly:
Uses decomp to get rid of the flail calculation's ugly decimals in favor of how the game actually calculates it.


UTDZac:
The purpose of this change is to match the calculation to as true as possible to the original code for calculating the values.

https://github.com/pret/pokefirered/blob/49ea462d7f421e75a76b25d7e85c92494c0a9798/src/battle_script_commands.c#L7929-L7942

UTDZac (approved):
This fix looks accurate to me. Good find.

PR #156, opened 2022-08-08
Provide an option to enable/disable the Last Damage calculation information

UTDZac:
There's been a bit of conversation about whether the Last Damage information is "too much" for the tracker to be doing. While most voices have said it's probably fine given you can just calc it yourself, I'd still like to include this option for those that don't want so much hand-holding.

This PR exposes an option to the user to turn the Last Damage calculation information on or off. When turned off, the damage carousel will simply be skipped (not visible).


Fellshadow (approved):
Agree that it's fine in most scenarios, but it's nice to have the option for those that don't want it

PR #157, opened 2022-08-08
Use stat judge messages, add a clear data button, and some typos/bug fixes

UTDZac:
- Correcting two ability names that have been inaccurate for a while now
- Changing IV estimate output message to match Stat Judge descriptions
- Adding a button to Clear the currently loaded Tracked Data, for times when it can bug out.
- Fixing bug with STAB color still showing for randomized move types when "reveal info" option is disabled


Fellshadow (approved):
nice fixes! Could maybe turn the Joey meme into an easter egg if their pokemon is rattata? 👀

PR #158, opened 2022-08-08
Water Spout / Eruption errors out; adding better error handling

UTDZac:
One of the updates as part of the #154 setup the calculations for Water Spout and Eruption to use variable power, instead of ">HP". This PR fixes that error.

`NLua.Exceptions.LuaScriptException: ironmon_tracker/Utils.lua:320: attempt to perform arithmetic on a nil value`

Also including some updates to the Main Tracker script for better error handling.


UTDZac:
Ninjafriend discovered a bug with the new ability read information I added, included that fix in here as well.

Fellshadow (approved):

PR #159, opened 2022-08-08
Improved version/file checks

Fellshadow:
Finally got round to future-proofing the Bizhawk version check. The updated check uses regex to check the major and minor version numbers separately, rather than the hardcoded string checks done before. This would then allow a theoretical 2.10 release of Bizhawk to pass the check.

Also, improved the error handling and loading of required tracker files. It only checked the Inifile.lua so would technically pass the check if for whatever reason a user moved any of the other files out the folder except the Inifile.lua. I've defined the files to load in a table (which can be edited if we add to/change the files in the data folder) so that a check is ran for each file it attempts to load before loading it.

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -40,7 +40,7 @@ function Main.Initialize()
 	print("\nIronmon-Tracker (Gen 3): v" .. Main.TrackerVersion)
 
 	-- Check the version of BizHawk that is running
-	if Main.UnsupportedBizhawkVersion() then
+	if not Main.SupportedBizhawkVersion() then
 		print("This version of BizHawk is not supported for use with the Tracker.\nPlease update to version 2.8 or higher.")
 		Main.DisplayError("This version of BizHawk is not supported for use with the Tracker.\n\nPlease update to version 2.8 or higher.")
 		return false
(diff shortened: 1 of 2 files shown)


Ironmon-Tracker.lua
```diff
@@ -11,55 +11,80 @@ function Main.Initialize()
...
+-- Returns true if Bizhawk version is older than 2.8
+function Main.UnsupportedBizhawkVersion()
+	-- Significantly older Bizhawk versions don't have a client.getversion function
+	if client.getversion == nil then return true end
+
+	-- Check the major and minor version numbers separately, to account for versions such as "2.10"
+	local major, minor = string.match(client.getversion(), "(%d+)%.(%d+)")
+	if major == nil or minor == nil then return true end
+
+	return (tonumber(major) < 2 or tonumber(minor) < 8)
```
UTDZac, line 74: A few possibilities here that might cause this to fail:
- "Bizhawk 3" (unlikely but still possible)
- "Bizhawk 3.1" (because `minor < 8`)

And super small, minor request, but could this function be inverted to "SupportedBizhawkVersion()" and the return results flipped? I think it just reads a little nicer that way.

UTDZac (approved):
Awesome, looks great!

PR #160, opened 2022-08-08
Restore fishing law

Fellshadow:
Instead of writing to memory to reset the item stored in `gSpecialVar_ItemId` to prevent a non-fishing encounter from being counted as a fishing encounter, can read the in-game tally of fishing encounters stored in the [game stats](https://github.com/pret/pokefirered/blob/master/include/constants/game_stat.h) to check if one occurred.

Also cleaned up the other game stats reads as part of this, moving the actual read to a Utils function and defining enums for the game stats currently being used in Constants.

To note, I set the fishing encounters tally in Tracker.Data to initialise to whatever the memory read of the in-game fishing encounters stat is at time of tracker data reset. This is to prevent a wild encounter being incorrectly tracked as a fishing encounter if someone were to clear their tracker data mid-run while the fishing rod is still set in `gSpecialVar_ItemId` in game memory.

Law and order has been restored ⚖️

### DIFF
--- a/ironmon_tracker/Constants.lua
+++ b/ironmon_tracker/Constants.lua
@@ -69,6 +69,13 @@ Constants.MoveTypeColors = {
 	unknown = 0xFF68A090, -- For the "Curse" move in Gen 2 - 4
 }
 
+Constants.GAME_STATS = { -- Enums for in-game stats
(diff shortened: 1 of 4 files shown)


UTDZac (approved):
Excellent find. Glad we can now continue to say the Tracker upholds its two primary rules since it no longer writes data to game memory.

PR #164, opened 2022-08-09
Levitate bug

jtigues:
- gBattleCommunication added for all games. Logs
- Using gBattleCommunication[MISS_TYPE] = B_MSG_GROUND_MISS to more accurately track Levitate
- Removed Levitate from the abilities groupings. There isn't a message logged for it so it should be handled outside of the structure that looks for messages
- Leaving gMoveFlags definitions in place, need it for the ability swapping detection coming later, and other tracking might care about checking for misses/failures

### DIFF
--- a/ironmon_tracker/Program.lua
+++ b/ironmon_tracker/Program.lua
@@ -458,6 +458,12 @@ function Program.autoTrackAbilitiesCheck(battleMsg, enemyAbility, playerAbility)
 		Program.SyncData.battlerTarget = -1
 	end
 
+	--Levitate doesn't get a message, so it gets checked independently
+	if enemyAbility == 26 and attacker % 2 == 0 and levitateCheck == 4 then
+		-- Requires checking gBattleCommunication for the B_MSG_GROUND_MISS flag (4)
+		return true
+	end
+	
 	-- Abilities to check via battler read
 	local battlerMsg = GameSettings.ABILITIES.BATTLER[battleMsg]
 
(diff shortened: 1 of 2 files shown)


ironmon_tracker/Program.lua
```diff
@@ -448,8 +448,8 @@ function Program.autoTrackAbilitiesCheck(battleMsg, enemyAbility, playerAbility)
 	local battler = Memory.readbyte(GameSettings.gBattleScriptingBattler) -- 0 or 2 if player, 1 or 3 if enemy
 	local attacker = Memory.readbyte(GameSettings.gBattlerAttacker) -- 0 or 2 if player, 1 or 3 if enemy
 	local battlerTarget = Memory.readbyte(GameSettings.gBattlerTarget)
-	local moveFlags = Memory.readbyte (GameSettings.gMoveResultFlags)
 	local currentTurn = Memory.readbyte(GameSettings.gBattleResults + 0x13)
+	local levitateCheck = Memory.readbyte(GameSettings.gBattleCommunication + 0x6)
```
UTDZac, line 452: Can you wrap this inside of the if-block that checks if the enemyAbility is Levitate? This way, we don't have to do an extra memory read every call to this function, and only do it while it's possible that the enemy has this ability.

Something like:
`if ability is 26 and attacker is us then... read from memory, if check, then...`

jtigues: Sure, done

UTDZac (approved):
Looks great.

PR #168, opened 2022-08-11
debug form that lets you edit pokemon on the fly

thisisatest1900:
to enable the form you need to enter Debug.createEditPokeForm()
in the lua console 
this creates a form that lets you change pokemon specie move set and chose between ability 1 or 2 
image

few things to note it lets you edit pokemon in your/opponent team from slot 1-6 
if you are editing pokemon that doesn't exist it will create it as dead pokemon cause i haven't edited stats yet cause its a bit more complicated

if you are in battle doing editing for example editing opponent poke for the whole changes to take effect you need to change the pokemon so use roar to switch the opponent poke is good option or to edit the second mon in the party and kill the first mon

ability editing lets you chose between ability 1 or 2 of the pokemon i didnt find a way to hard choose the ability

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -1,4 +1,4 @@
-Main = { TrackerVersion = "0.6.0a" } -- The latest version of the tracker. Should be updated with each PR.
+﻿Main = { TrackerVersion = "0.6.0a" } -- The latest version of the tracker. Should be updated with each PR.
 
 Main.CreditsList = { -- based on the PokemonBizhawkLua project by MKDasher
 	CreatedBy = "Besteon",
@@ -35,6 +35,7 @@ function Main.Initialize()
 		"/Program.lua",
 		"/Pickle.lua",
 		"/Tracker.lua",
+		"/Debug/Debug.lua"
 	}
 
 	print("\nIronmon-Tracker (Gen 3): v" .. Main.TrackerVersion)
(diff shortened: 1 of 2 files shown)


Fellshadow (changes requested):
Nice, certainly a helpful feature! Just wanna rework about how this is handled, appreciate that you've already made it a separate file.

Ideally we want to keep this out of main release, so i'd ask that you remove it from loading with the rest of the tracker files and instead rework Debug.lua to act as a separate script that a dev could load into the lua console alongside the tracker for Debug options
  - Can have it run `Debug.createEditPokeForm()` when it starts up (if that's the only function the dev would use) so the options are available to a dev using it. If this file gets expanded in the future to add other debug options then can have it print out the available commands for a dev to use or something
  - This way, we can just exclude the Debug.lua file from releases so it's not available to end-users

Also to note, we're changing the flow of things on the main repo (not your fault, it's literally only just happened lol). Planning to write up a section about this in documentation but the short of it is that we wanna merge new features / updates to features to staging first, and then when the build is tested and ready for a new release it'll be pushed to main. So i'm gonna change this PR to go into staging, just so you know

thisisatest1900:
changed how to load the file since it relays on data that tracker loads 
the order to open it is to have tracker loaded then open the debug script
if you load it before there is print message that ask you to load tracker before 
after the initial load you can close the tracker if you want i guess i just didn't want to do all the loading in my script
because if there would be changes in the future to how its loaded it would break my script

Fellshadow (approved):
Yeah that makes sense to do it that way, since it's a debug option for the tracker it *should* only be loaded with the tracker anyways but always good to have that safety check for the order of loading.

Currently no plans to change how the tracker loads in a huge way as far as I know, but if things do change this could be adapted alongside whatever rework is done accordingly

PR #170, opened 2022-08-11
New Feature: Animated Pokémon Popout Add-on

### LINKED ISSUE #167: Feature: Animated Pokemon Sprites as a popout window
Creating this as a new Issue / Feature Request to track progress and post sources.

This feature will display your lead Pokemon as an animated sprite/image in a popout window. The user can then put this cool image on their stream to show off their Pokemon.

Source images: https://projectpokemon.org/home/docs/spriteindex_148/3d-models-generation-1-pok%C3%A9mon-r90/

Working branch: https://github.com/besteon/Ironmon-Tracker/tree/user/utdzac/animated-pokemon-popout

I have most of this working at the moment in a new branch. One question I have is if having all of these images takes up a ton of space on disk (combined this is very likely), then we want to find a way to release this as an add-on somehow. I'm thinking the main project can include all of the code and options for this, but will inform the user they need to download the expansion pack of images to properly use it.

UTDZac:
resolves #167 

This is a feature "Add-on" to the Tracker. It works a little bit differently than a normal feature addition in that the feature itself is not fully available in the main Tracker application. The user can see that it exists, but the feature is disabled.

To enable the feature for use, the user will have to install the additional files required by the add-on, in this case, all of the animated images. This "installation" is a simple process of downloading a zip of all of the animated files, then placing them in the correct (empty) folder within the Tracker files. The user then refreshes the Tracker script and the feature will be enabled and functional.

The add-on files for this feature will be available through our Wiki pages. For future feature add-ons like this, ones that require a significant amount of space on disk, we can use a similar process of separating a feature from its required files.

The animated Pokemon popout window can be moved anywhere, and added to a user's OBS as a source, to nicely overlay on top of their stream. The popout uses a transparency key of "Magenta" by default. Currently the only way to change this is through `Constants.lua`

image


UTDZac:
Documentation for this feature can be found here: https://github.com/besteon/Ironmon-Tracker/wiki/Tracker-Add-ons

Only thing lacking is a place to put the 50MB~ download for the animated GIFs

Fellshadow (approved):
i'm sure a lot of people will love this!
I saw you already figured it out, but fwiw i agree that probably the best way to link to the additional folder download is adding the pokemonAnimated folder as a separate asset on next release then link to that asset in the wiki

PR #172, opened 2022-08-12
Additional versions

jtigues:
- Perish Song fixed for Soundproof
- Still missing Ruby/Sapphire, and foreign FR. checking through decomps atm for the former


UTDZac (approved):
Looks solid. There's still a debug message in here, but that's fine given you or someone else may want to exploring solving this for the other games.

PR #173, opened 2022-08-12
Bugfix/soundproof

jtigues:
- Ruby, Sapphire, and other languages of Fire Red


UTDZac (approved):
Awesome, thanks for doing this.

PR #179, opened 2022-08-13
Color change

jtigues:
- Move Effectiveness and STAB is now properly tracked as the pokemon's types change in battle
- Types shown in battle will also change 
- Color Change, Camoflauge, Conversion, Conversion2, Transform
- Couple minor release related changes from main

https://user-images.githubusercontent.com/12836386/184515480-7320857e-2a34-43e8-ae04-4142c720711c.mp4


UTDZac (approved):
Thanks for doing this and setting the stage for handling dynamic battle data as a temporary data set. Looks great except for a small edge case that can't actually occur because of safety checks, but still worth fixing for anyone else that might use this code.

PR #187, opened 2022-08-15
Info screen notepad

jtigues:
- Note added to the bottom of the Info Screen
- minor QOL adjustments to the behavior of the note taking functions
- currently sharing TrackerScreen.openNotePadWindow between both buttons. Maybe it should move to Tracker.lua? Doesn't cause any problems as is

image

### DIFF
--- a/ironmon_tracker/Tracker.lua
+++ b/ironmon_tracker/Tracker.lua
@@ -341,8 +341,8 @@ end
 -- If the Pokemon is being tracked, return its note; otherwise default note value = ""
 function Tracker.getNote(pokemonID)
 	local trackedPokemon = Tracker.getOrCreateTrackedPokemon(pokemonID)
-	if trackedPokemon.note == nil then
-		return ""
+	if trackedPokemon == nil or trackedPokemon.note == nil then
+			return ""
 	else
 		return trackedPokemon.note
 	end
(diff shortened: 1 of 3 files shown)


ironmon_tracker/Tracker.lua
```diff
@@ -341,8 +341,8 @@ end
 -- If the Pokemon is being tracked, return its note; otherwise default note value = ""
 function Tracker.getNote(pokemonID)
 	local trackedPokemon = Tracker.getOrCreateTrackedPokemon(pokemonID)
-	if trackedPokemon.note == nil then
-		return ""
+	if trackedPokemon == nil or trackedPokemon.note == nil then
```
UTDZac, line 344: Up to you if you want to revert this, but the `Tracker.getOrCreateTrackedPokemon(pokemonID)` will never return `nil` by design, since it either gets the current Pokemon or creates a new one if one doesn't exist; it always returns something that can be tracked.

UTDZac (approved):
I agree with re-using the same open notepad dialog box for this feature. Ideally we can have it in a more central location since it is now shared between multiple screens. Maybe other popups can be moved too, such as a `Forms.lua` which can give us better control over what's where.

This is also a good idea because I expect us wanting to merge the ability prompts into the notepad prompts, to accommodate the upcoming ability info screen.

Issue #182, opened 2022-08-14
Route Info: The MORE button shouldn't be visible when there are no other encounter tables


UTDZac:
Version: 0.6.1

Currently the More button is displayed on the Route Info page at all times, even for zones that only have a single encounter table. This can potentially be confusing as the button exists but nothing happens when it is pressed.

Ideally this button should be hidden if no other encounter table is visible. This might not be an easy fix, as we eventually want Trainer encounter tables added in. So we'd need a way to properly distinguish this, or consider how a user would want to use this information.

![image](https://user-images.githubusercontent.com/4258818/184551214-df8caf82-3921-4aa2-99b5-cd2ec17e4fa7.png)

Fellshadow:
I wonder if there'd be enough space to implement back/forward buttons like this to cycle between encounter pages, which would be much clearer imo for the intended purpose. These wouldn't necessarily need to be removed if there aren't more pages to view due to being clearer that it's to cycle encounter types but of course would still be cleanest/ideal to do so
Idea mockup:
![Screenshot_20](https://user-images.githubusercontent.com/106463662/184627082-9d945e23-364a-433b-8a72-39f3df9971d8.png)

PR #188, opened 2022-08-15
Replace route info MORE button with back/forward cycle buttons

Fellshadow:
Alternative solution for #182, replaces the MORE button with a back and forward button to cycle through encounter tables.

The original issue was to hide the MORE button when there are no other encounter tables, due to it being potentially confusing. I feel like back/forward buttons are clear enough to not have to hide them for no other encounter tables / the header text would look out of place on its own without them but I'd be happy to go back and rework this a little to also hide them when no other encounter tables are there if that is desired

I had to change the header text for static encounters to "Static Pokémon encounters" to make it fit.

Demo:
RouteInfoButtons

### DIFF
--- a/ironmon_tracker/data/RouteData.lua
+++ b/ironmon_tracker/data/RouteData.lua
@@ -215,6 +215,31 @@ function RouteData.getNextAvailableEncounterArea(mapId, encounterArea)
 	return encounterArea
 end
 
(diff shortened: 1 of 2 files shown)


UTDZac (approved):
Simple yet great change. Thank you. I like the idea of using the arrows to navigate the different encounter tables.

PR #189, opened 2022-08-15
Add safari zone data

### LINKED ISSUE #181: Route Info: Safari zone data missing from FRLG and RSE
Version: 0.6.1

Currently the Safari Zone wild encounter data is not available for all games. This is simply a matter of getting it manually entered in. I hadn't done this original because the Safari Zone isn't really used in Ironmon.

FRLG:
> https://github.com/besteon/Ironmon-Tracker/blob/main/ironmon_tracker/data/RouteData.lua#L1039-L1042

RSE:
> #define LAYOUT_SAFARI_ZONE_NORTHWEST 238
> #define LAYOUT_SAFARI_ZONE_NORTH 239
> #define LAYOUT_SAFARI_ZONE_SOUTHWEST 240
> #define LAYOUT_SAFARI_ZONE_SOUTH 241

Fellshadow:
Resolves #181 - decided to also add the additional two safari zone areas unlocked in Emerald post-game as they were on the same Pokéarth page and it wasn't much hassle

For FRLG I found the full name fit on the tracker enemy pokemon view, not sure about RSE since i don't have a save file to test that with (and there's areas like "North-West" there) so left it as abbreviations


UTDZac:
Thank you for tackling this. I was intimidated at the size of these areas when initially adding in all the other route data. It's so much data.

PR #193, opened 2022-08-17
Separating Battle logic from Program.lua

UTDZac:
Initial effort of separating out a ton of battle logic from `Program.lua` into a new file, `Battle.lua`. Wanting to get this merged in sooner rather than later (and update `bugfixes` branch afterwards), so others can build off of it.

The main goal is separating Program/Battle based on the responsibilities of each: 
- **Program** likely handles anything related to the frame loops, the timings for handling input and drawing. It also does everything that involves reading data from memory (a few loose ends in Drawing/Utils). 
- **Battle** manages some of the Programs data at a more granular level, and tracks any entire fight. This allows better data management for when in a battle, more accurate data reads, and condensing some of the redundant code from a few of our recent features. (granted, not many improvements right now, but its a start)

Eventually we can have `Battle.lua` provide us with a simplified set of data to draw to the screen. Currently all of this logic is scattered throughout `TrackerScreen.drawMovesArea(...)` which has to perform all of the "calculations" inside a "drawing" function. These ideally would be separated.

Also worth noting, we should probably test this a good bit, especially with some of the ability triggers, and very especially synchronize.

### DIFF
--- a/ironmon_tracker/data/MoveData.lua
+++ b/ironmon_tracker/data/MoveData.lua
@@ -204,7 +204,7 @@ function MoveData.moveHasUnusualPower(moveId)
 	return false
 end
 
-function MoveData.isValidMove(moveId)
+function MoveData.isValid(moveId)
 	return moveId ~= nil and moveId >= 1 and moveId <= MoveData.totalMoves
 end
 
--- a/ironmon_tracker/data/MoveData.lua
+++ b/ironmon_tracker/data/MoveData.lua
@@ -1,4 +1,6 @@
-MoveData = {}
+MoveData = {
+	totalMoves = 354,
+}
 
 -- Move categories identify the type of attack a move is: physical, special, or status
 MoveData.Categories = {
@@ -96,7 +98,7 @@ function MoveData.readDataFromMemory()
 	-- If any data at all was randomized, read in full move data from memory
 	if MoveData.checkIfDataIsRandomized() then
 		print("Randomized move data detected, reading from game memory...")
-		for moveId=1, #MoveData.Moves, 1 do
+		for moveId=1, MoveData.totalMoves, 1 do
 			local moveData = MoveData.Moves[moveId]
 
 			moveInfo = MoveData.readMoveInfoFromMemory(moveId)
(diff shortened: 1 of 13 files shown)


ironmon_tracker/data/MoveData.lua
```diff
@@ -1,4 +1,6 @@
-MoveData = {}
+MoveData = {
+	totalMoves = 354,
```
Fellshadow, line 2: any particular reason to move the length to a hardcoded variable instead of using # to get the length?

I don't forsee us adding/removing moves/pokemon to be fair so it's not an issue, but if we were to then would also need to remember to update the totalMoves/totalPokemon variables. I'm just curious as to why you felt this change was needed

Fellshadow (commented):
Just a couple things from looking at the code that i've mentioned but other than that i'm happy for it to go in, though i'm gonna do some testing first of at least some ability stuff and what-not. Will update here how that goes

Also there was a big breaking bug introduced in recent bugfixes, i've put through a fix on bugfixes/staging so may wanna update this branch with it too (though there's no conflict here from it so it's probably fine regardless, just noting it)

Fellshadow:
Update, did some brief ability testing and it seemed to work fine. Don't have the means to test a lot of edge-cases like synchronize going off from your static but brief synchronize testing from proccing with moves worked fine

ironmon_tracker/data/MoveData.lua
UTDZac: I want to say I opted to do this because using `#` to get the size of the table might perform a calculation that has to determine how big the table is, then return that value; as opposed to us just setting the value here and using it. Maybe with Lua that is not the case.

ironmon_tracker/screens/TrackerScreen.lua
```diff
@@ -632,16 +634,15 @@ function TrackerScreen.drawPokemonInfoArea(pokemon)
 			Drawing.drawButton(TrackerScreen.Buttons.PCHealAutoTracking, shadowcolor)
 		end
 	else
-		if Tracker.Data.trainerID ~= nil and pokemon.trainerID ~= nil then
+		-- Assumes we are in Battle, unsure if this check is neccessary
```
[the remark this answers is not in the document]
UTDZac: Good catch and smart habit to avoid assuming anything. I suppose what I was thinking here was eventually we'd extend the route info stuff to work outside of battle, mostly for mapping trainers.

Fellshadow (approved):
ye looks good to me

PR #196, opened 2022-08-17
Bill's Island Adventure! (Sevii Isles Data)

### LINKED ISSUE #191: Route Info: Some battles do not have a mapId or route name
Version: 0.6.1

After playing for a bit and watching others play through the game, encountering pokemon from wild battles or trainer battles, there are a handful of areas I've noticed that show up as blank.

**[FRLG]** Areas that the player can engage in battle that have no route info (no route name/mapid):
- Each individual room within the S.S. Anne; outside areas appear to be fine
- The Game Corner rocket grunt at the entrance [Map Id = 27]

I suspect there may be more places that the player can engage in battle that show a blank route name, both in FRLG and RSE. Especially Sevii Islands.

Adding these might be a bit tricky. If I had a save file at these spots, then I could more easily just output the mapId to make sure we get the right one. Otherwise, I used this table of definitions for map stuff: https://github.com/pret/pokefirered/blob/918ed2d31eeeb036230d0912cc2527b83788bc85/include/constants/layouts.h

Fellshadow:
Resolves #191 
- Adds a bunch of data for the Sevii Isles routes/locations for wild pokemon (and empty encounter tables for future trainer info)
- Adds a check to distinguish rock smash encounters using `GAME_STATS_USED_ROCK_SMASH` (to tell a rock has been smashed) and `gSpecialVar_Result` (which stores the result of the wild encounter roll)
- Adds in map IDs for Celadon Game Corner (for the rocket grunt) and the rooms on-board the SS Anne

### DIFF
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -202,7 +202,7 @@ function Battle.checkEnemyEncounter(opposingPokemon)
 
 		local rockSmashResult = Memory.readword(GameSettings.gSpecialVar_Result)
 		if rockSmashResult == 1 then
-			Program.CurrentRoute.encounterArea = RouteData.EncounterArea.ROCKSMASH
+			Battle.CurrentRoute.encounterArea = RouteData.EncounterArea.ROCKSMASH
 		end
 	end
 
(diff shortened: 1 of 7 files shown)


UTDZac (approved):
Really appreciate you doing this. Great addition!

PR #198, opened 2022-08-18
Can now keep theme presets in a file

UTDZac:
Allows users to keep their favorite theme string codes in a `ThemePresets.txt` file, for easier theme management.

Updating the version number so that I can have some people test this.


Fellshadow (approved):
Looks good, will be cool to be able to add to the presets drop-down in an easier way for sure.

Only question I have really is the longer-term plans for building out this functionality / minimising user error. Since the way to add a theme to the tracker list is by editing the .txt file, there's the obvious concern is what if people wind up opening it up and adding in garbage values, deleting random parts of it, or anything along the lines of editing it in a way that's not expected / doesn't make much sense to do? (If it *can* happen then *someone* will do it as we know lol)

I know there's safety checks throughout from the existing import functionality which would presumably catch these and at least prevent a tracker crash. My guess is that if something isn't right with a particular theme string in the .txt file it either doesn't get included at all in the presets dropdown or if it is "correct" enough to pass that check it'll just fail to load when selecting from the drop-down? Not a big issue at the moment (as long as it doesn't cause crashes of course) but certainly stuff to think about for improving the safety nets down the road

UTDZac:
Agree with adding better support for helping the user understand what may have gone wrong with the theme they just added. Having a bunch of errors get spat out in their face or popups probably isn't the best experience.

Also agree that long-term we could have something more user-friendly for selecting a preset and/or editing it live, without requiring them to export the code and append it. We'll have to see as more and more people use this, and as we get more feedback or requests for a better user experience.

PR #199, opened 2022-08-19
add new game check for trainer id

### LINKED ISSUE #186: Encounter data fails sometimes when a new game starts
Version: 0.6.1

Under some rare conditions that have to do with restarting the current ROM with Bizhawk, the old tracker data from the previous run will be loaded into the new run. This is problematic because as a new game begins the Trainer ID of the player will change. The Trainer ID is used to determine if an encounter with a Pokémon is a wild encounter ("Trainer ID" of the wild matches the user) or a battle with a trainer (mismatched IDs).

The current code determines the Trainer ID of the player simply by waiting until they receive their first Pokemon, then setting the Trainer ID in `Tracker.Data` to that value.

Note: All wild Pokémon have a Trainer ID that matches the player. All gift Pokémon and Pokémon belonging to an enemy trainer have a different Trainer ID.

image

**References:**
- Trainer ID read in from here <https://github.com/mkdasher/PokemonBizhawkLua/blob/aef8c355bb0e0c69f00f5edb656722787c97f5e0/pkmgen3/Program.lua#L77>
- and here <https://github.com/mkdasher/PokemonBizhawkLua/blob/aef8c355bb0e0c69f00f5edb656722787c97f5e0/pkmgen3/GameSettings.lua#L44>

**Known Issues/Blocks:**
- Implementing this fix seems particularly challenging if we want to do it in such a way as to not repeatedly read in the Trainer ID from memory all the time. Ideally the ID is read in exactly once and doesn't have to be read from memory again. However, I don't know how to *know* for certain the ID that is currently stored in Tracked Data is accurate to the current game or if its the wrong ID and needs to be read in again.

Fellshadow:
Addition to help with #186 (if not a full solution)

Adds in a new game flag for checking trainer ID. The linked issue is usually due to reloading a rom that has the same rom hash as the one that was already open (re-randomising the seed changes this rom hash usually from what i saw) so this mainly covers if you reload the current rom to begin a new game (for example maybe you want a new vanilla playthrough instead)

The process by which this works is as follows:

- Store a `isNewGame` flag in `Tracker.Data` and initialise to true by default if no tracker data loaded (i.e. rom hash alone has determined that it's a different rom)
- If `isNewGame` is false then check if player currently has any pokemon in the current party (i.e. if there's any pokemon in `Tracker.Data.ownTeam`)
  - If there are no pokemon in the current party then set `isNewGame` to true
- If `isNewGame` is true when the player has a valid pokemon, then read the trainer id as we currently do
  - If the trainer id in the current tracker data is 0 or nil (from data reset) then set the newly read trainer id as the trainer id as is currently done
  - If the newly read trainer id is different to the one stored currently in `Tracker.Data` (i.e. old data has been loaded in) then reset the tracker data. After the reset then set the trainer id to the newly read id.
- Once we set a trainer id from the last step, set `isNewGame` to false

This just expands upon the existing trainer id check to set a trainer id so no additional memory reads

The only time this should reset the tracker data outside of how it's currently being reset is if the trainer id we read is different to the one that's currently stored while the `isNewGame` flag is set to true, so should avoid accidental deletion of data.

EDIT: a possible scenario in which this would fail is if the tracker is loaded up mid-game with no tracker data to load, and the pokemon in the first slot of the party is a traded pokemon with a different trainer id. As such the better workaround may be to just rework [reading trainer id from `gSaveBlock2`](https://github.com/besteon/Ironmon-Tracker/issues/186#issuecomment-1214878898) instead in a way that minimises how much we make the reads / minimises performance impact

### DIFF
--- a/ironmon_tracker/Program.lua
+++ b/ironmon_tracker/Program.lua
@@ -142,6 +142,12 @@ end
 function Program.updatePokemonTeams()
 	-- Check for updates to each pokemon team
 	local addressOffset = 0
+
+	-- Check if it's a new game (no Pokémon yet)
+	if not Tracker.Data.isNewGame and Tracker.Data.ownTeam[1] == 0 then
+		Tracker.Data.isNewGame = true
+	end
+
 	for i = 1, 6, 1 do
 		-- Lookup information on the player's Pokemon first
 		local personality = Memory.readdword(GameSettings.pstats + addressOffset)
@@ -152,11 +158,22 @@ function Program.updatePokemonTeams()
 
 			if Program.validPokemonData(newPokemonData) then
 				-- Sets the player's trainerID as soon as they get their first Pokemon
-				if Tracker.Data.trainerID == nil or Tracker.Data.trainerID == 0 then
-					Tracker.Data.trainerID = newPokemonData.trainerID
+				if Tracker.Data.isNewGame and (newPokemonData.trainerID ~= nil or newPokemonData.trainerID ~= 0) then
+					if Tracker.Data.trainerID == nil or Tracker.Data.trainerID == 0 then
+						Tracker.Data.trainerID = newPokemonData.trainerID
+					elseif Tracker.Data.trainerID ~= newPokemonData.trainerID then
+						-- Reset the tracker data as old data was loaded and we have a different trainerID now
+						print("Old/Incorrect data was detected for this ROM. Initializing new data.")
+						Tracker.resetData()
+						Tracker.Data.trainerID = newPokemonData.trainerID
+					end
+
+					-- Remove trainerID value from the pokemon data itself since it's now owned by the player, saves data space
+					newPokemonData.trainerID = nil
+
+					-- Unset the new game flag
+					Tracker.Data.isNewGame = false
 				end
-				-- Remove trainerID value from the pokemon data itself since it's now owned by the player, saves data space
-				newPokemonData.trainerID = nil
 
 				Tracker.addUpdatePokemon(newPokemonData, personality, true)
 			end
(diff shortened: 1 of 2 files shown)


ironmon_tracker/Program.lua
```diff
@@ -152,11 +158,22 @@ function Program.updatePokemonTeams()
 
 			if Program.validPokemonData(newPokemonData) then
 				-- Sets the player's trainerID as soon as they get their first Pokemon
-				if Tracker.Data.trainerID == nil or Tracker.Data.trainerID == 0 then
-					Tracker.Data.trainerID = newPokemonData.trainerID
+				if Tracker.Data.isNewGame and (newPokemonData.trainerID ~= nil or newPokemonData.trainerID ~= 0) then
```
UTDZac, line 161: The `(newPokemonData.trainerID ~= nil or newPokemonData.trainerID ~= 0)` part seems unneeded. Mostly because we just passed the `validPokemonData()` check, so I'd expect this to be a real Pokemon; real data means a real TrainerID.

Either way, the OR condition kind of fails here to serve anypurpose, because if `trainerID` is `0` for example, it will pass the first half of the check, and if its `nil` it will pass the second half.

UTDZac (commented):
I had looked over this briefly nearly 12 hours ago and remember seeing something that stood out to me as weird, or a potential edge case that could break this. I'm trying to remember again at 2am what that might have been, but just can't. Given that, I'm going to hold off on fully approving or merging this for a bit.

With that said however, I do want something like this included before the next release goes out.

UTDZac (approved):
Went ahead and made the two requested changes since they were fairly quick. Still unsure about the weird edge-case I was thinking about many days ago, maybe it was nothing after all. Am willing to try this solution for 0.6.2 and see if it fixes most occurrences of the core issue.

UTDZac:
Did some lite testing this evening. Had a situation where I restarted the game and the tracker still showed old pokemon from my previous run. The tracker correctly identified that I had old data and replace the trainer id. It worked perfectly after that.

PR #200, opened 2022-08-20
Focus Punch

jtigues:
- Looks like after the Battle/Program file separations, the Focus Punch logic was checking if the opponent was attacking using the attacker from the last damage calculation. Since the charge up part of focus punch doesn't deal damage, it was getting skipped. Updated to use the actual attacker flag


UTDZac (approved):
Great find, thats for looking into this

PR #205, opened 2022-08-22
Feature: Adding option to disable carousel on the main Tracker screen

UTDZac:
There have been a few requests to be able to further disable the carousel from changing the screen. Notably, the main tracker screen with the badges. Any movement there is very distracting. This feature includes an option for that.

I've also removed startup tips since, well, after several users tested it, they weren't really doing what I had hoped.


UTDZac (commented):
Just double-checking my changes here on a new day. It's a fairly minor fix, so approving this one.

PR #206, opened 2022-08-22
Ability info screen

jtigues:
- AbilityData.lua set up for abilities, general and Emerald-specific descriptions added. References to the old Abilties list updated accordingly.
- Ability tracking moved to the same form as Notes
- Ability Info Screen added; clicking an ability brings up the Ability Info Screen.
- Abilities can be searched, and will display a name + description + the emerald-specific description, if applicable


UTDZac (approved):
Lots of changes here but everything looks to be in order. Thank you for tackling this one. I reviewed most everything except `TrackerScreen.openNotePadWindow()` as it's quite lengthy now and I really just want to open it up and play with it some.

I also want to more thoroughly review all of the ability descriptions and tidy them up a bit where needed, if any. But I can get to that in the near future.

PR #207, opened 2022-08-22
add Main.GenerateNextRom logic

davidhouweling:
## Context

This PR extends on the work done by @UTDZac by adding logic to generate the rom on the fly.

## How has this been tested

Manually using a Pokemon Fire Red ROM, verified that it generates successfully and loads up appropriately
- Follow the UI to set up all the necessary paths needed for the generator
- Force a reset with A+B+Sel+Start
- Check the Lua logs for the output and play game

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -248,17 +248,66 @@ function Main.GetNextRomFromFolder()
 end
 
 function Main.GenerateNextRom()
-	-- TODO: Implement this as part of #203
+	if not Main.FileExists(Options.FILES["Randomizer JAR"]) then
+		print("Randomizer Jar file not found.")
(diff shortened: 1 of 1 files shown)


UTDZac (approved):
Thank you so much for this sweet addition to the project. I look forward to trying it out.

With regards to handling the way the script generates a new ROM file in time for it to be used when the user activates the Quick-load feature, I agree that we probably want to create this file ahead of time and have it sitting in waiting. I'll explore some options here.

A few minor notes with regards to future additions to the project (no action required on your part for this PR):
- It looks like some of the tab formatting on your end was using four spaces instead of an actual TAB insert.
- When able, try and use the Lua string libraries for string operations instead of `bizstring`. In the distant future we may be porting this project to mGBA which doesn't have access to Bizhawk only things.

PR #210, opened 2022-08-23
Ability info screen

jtigues:
- Back button colors fixed
- spacing adjustments
- added example to Compoundeyes
- added clarification text for gaining the ability curing the status for Limber/Vital Spirit/Insomnia/Water Veil/Magma Armor/Immunity/Own Tempo/Oblivious
- minor spelling/punctuation fixes
- Clear button no longer clears note text
- Accuracy and Evasion are capitalized now
- minor description syncing adjustments
- 'Emerald: ' color set to default
- Static/Poison Point/Flame Body/Cute Charm updated to 1/3 chance, since apparently that's what the actual logic in the decomp reflects. Might cause some confusion though since all wiki sources claim it's 30%


ironmon_tracker/screens/InfoScreen.lua
```diff
@@ -910,7 +909,9 @@ function InfoScreen.drawAbilityInfoScreen(abilityId)
 		end
 	end
 
-	Drawing.drawButton(InfoScreen.Buttons.close, boxInfoBotShadow)
+	local closeButton = InfoScreen.Buttons.close
+	closeButton.boxColors = { "Upper box border", "Upper box background" }
+	Drawing.drawButton(InfoScreen.Buttons.close, boxInfoTopShadow)
```
UTDZac, line 914: This permanently changes the color of the close button used for any info screen. The `closeButton` is still a reference to the original button.

A few solutions here are:
1. Temp store the colors, change them for the draw, then change back.
2. Use the bottom box color for the whole box, instead of top box color
3. Find a way to split the box into two, similar to Move Info Screen. We can likely use the top box for fun stats but I haven't thought about what those are.

I don't have a preference on which. I suppose at some point in the future we might end up on the third option.

UTDZac (approved):
Looks great, thank you for these updates.

PR #211, opened 2022-08-24
Finalizing the new quick-load feature

### LINKED ISSUE #203: Feature Request: Allow the tracker to create a new randomized seed on the fly
Version: 0.6.1

The most common use-case for the Tracker loading the next seed is for the user to independently create a batch of multiple ROMs, in sequential order, and put them all in a single folder. Once they have completed this, they can assign the folder to the Tracker through the settings. This functionality works great and should not be changed.

Many users like making a single randomized game at a time instead of multiple. Thus they don't use the quick-load feature, but also cannot benefit from the quick-load, or rather fast reset feature that the Tracker offers. This feature request is to allow for this (concept came from a streamer)

The largest benefit of this feature addition would mean that, once its all setup properly, the user doesn't ever have to interact with the Randomizer program to make seeds. They can simply press the quick-load buttons to just create a new seed and begin playing.

### Research
Determine if it's possible within LUA to create a randomized ROM using the randomizer. Likely possible through the CLI tool, assuming a settings file has already been created.

### Implementation
For now, I'm mostly interested in how possible this is and less interested in finding spacing within the tracker settings to expose this feature. There is likely a lot of setup required behind the scenes to get this to work, such as putting the required randomizer files in a location that the Tracker can find, and having a `.rnqs` settings file made and ready to go.

UTDZac:
Resolves #203 

This adds the final touches to the quick-load feature for generating a new rom on the fly.

**[NEW]** Includes the addition of an "Attempt counter" which is a simple text file that contains only a number. This can be used by streamers to show the number of attempts and allows formatting.

**[NEW]** Includes preloaded `.rnqs` files so the user can very quickly pick an Ironmon Ruleset to play with, instead of needed to get the settings string from the discord or rules page.

image


UTDZac (commented):
Doing a quick second review of the code. Also had Spaghetti quickly skim it as well.

PR #216, opened 2022-08-26
Evolution tracker delay

jtigues:
- New variables added for Evolution tracking
- Note: R/S have a slightly different setup for finding the Evolution task, and a different task state flow, so the other games have a section specific to them
- The Tracker will now no longer update Team Data while an Evolution task is active.
-Note 2: Foreign language Fire Red addresses are calculated according to similar addresses, but not tested directly. The 0x08 addresses have variable offsets relative to Fire Red 1.1, so it is not clear where Task_EvolutionScene actually is. If someone has access to the foreign versions, the memory address can be printed from the task itself during an evolution.

https://user-images.githubusercontent.com/12836386/186827770-11c56ddd-5bd1-4256-a880-fe92c1bbe64e.mp4


ironmon_tracker/Program.lua
```diff
@@ -437,6 +439,28 @@ function Program.isInCatchingTutorial()
 	return Program.inCatchingTutorial
 end
 
+function Program.isInEvolutionScene()
+	local taskLocation = GameSettings.sEvoStructPtr
+	--print (taskLocation)
+	if GameSettings.versioncolor ~= "Ruby" and GameSettings.versioncolor ~= "Sapphire" and GameSettings.versioncolor ~= "" and GameSettings.versioncolor ~= nil then
```
UTDZac, line 445: You can simplify this check by using the `GameSettings.game`, which has a value of `1` if the game is exactly Ruby or Sapphire, and some other number for the other games. You can also safely assume the values won't be `nil` or empty here, since those are required to be set before the Tracker will even start up.

So I'd recommend changing this to: `if GameSettings.game ~= 1 then`

UTDZac: And for clarity for the reader, I'd also invert the comment directly below this to be explain what we're looking at, since a `~= 1` check doesn't tell us much at a glance.

UTDZac (commented):
Thank you for digging into this really strange and rare issue, and putting together this fix.

Just a few minor things worth cleaning up here that I left notes on (see below).

In addition, I wonder if there's some sort of non-memory read condition we can check to avoid going through the entirety of this method when it's not going to matter 99% of the time. I spent a solid 5 minutes trying to think about various checks, but couldn't come up with anything solid.

jtigues:
Brainstorming at like 7 am:
- as a battle solution, we could disable updating when the pokemon's level is at or past the evolve condition and we are not in battle
- for out of battle, not sure if there's another alternative, since a lot of the other hooks are dependent on returning you to the battle flow
- thought about looking for the evolution music playing, but that only plays during the animation anyway, and we want the update delayed until after the move learning 
- maybe instead of updating while we are evolving,  we could update only while in the overworld. Might be fewer checks, but idk how much of thst info gets set after the battle ends and before the evo starts off the top of my head

UTDZac:
Thanks for the updates. Looks good.

PR #217, opened 2022-08-26
update vscode setup

davidhouweling:
This PR adds recommended VS code extensions, as well as sets up the workspace settings for the lua extension.


UTDZac:
Thank you for setting this up. Would you be able to walk me through exactly what this does. I'm inexperienced with this sort of thing and not sure what this does or what I'm expected to use it for.

Would be interested to here what this does for the project as well as how people are expected to use it.

davidhouweling:
Hey, for this PR i've added [VS Code recommendations](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions) for this workspace (i.e. this codebase). The 2 that i've set is EditorConfig and Lua checker (as there were already settings checked in specifically for this extension).

The other change in the `settings.json` was to remove the configuration prompt that comes up when you install the Lua checker extension to just localise the settings to the workspace.

UTDZac:
Would you be able to provide some tests cases such that I can try out this change? I'm still very unfamiliar with what is going on here. I believe this changes something about the project, but I don't know how to confirm that or test it myself.

I'd like to be able to apply this change, to test it to verify that it works, to review the added functionality and see if it's what we want, as well as offer any changes or recommendations before committing this to our project.

davidhouweling:
Hey, if you open up vscode and uninstall editorconfig extension, and your lua extension as well, then close and reopen vscode you should see a notification in the bottom right like the following...

image

And also, if you goto the extensions tab, under recommendations you'll see at the top of the list are the ones the workspace recommends (shown by a Star) which helps the developer install the ones this repo recommends (in this commit).

image

Why these two?
- Editorconfig: ensures that things like indentation is consistent (my default is 2 spaces, but this repo prefers tabs) without thought
- vscode-lua: this provides lint warning/errors in the IDE (a few came up which we can resolve later) and is the one suggested by the [luacheck github action](https://github.com/marketplace/actions/luacheck) which we could potentially add to this repo in the future which would help ensure lint warnings/errors are raised as part of PR requests (reducing the need for the reviewer to need to focus on these). Linters are great because it helps capture simple issues such as globals being used where they aren't expected, or if we go to the extent of typing variables/parameters, linters can ensure that their usage is correct as well (should it be passed a number or a string for example).

Specifically with vscode-lua, you'll see more useful guides to reduce potential for errors to be introduced.

image

Following up from this PR we should look to resolve the present linting issues, and then potentially introduce https://github.com/marketplace/actions/luacheck to ensure future code remains linted.

UTDZac:
Thank you for the screenshots. I feel like I'm still missing some steps here on getting this setup.

I've swapped my branch to this PR, I've uninstalled EditorConfig and reloaded VS Code. I do *not* see any recommendations pop-up. No matter, I just searched for the two that you listed and installed them. At this point I'm at a loss for how to initiate anything related to the lua check. I don't see any recommendations on my code in my IDE.

I'm sorry I'm really not great with any of this stuff, I have not used it before. Any step-by-step instructions on how to install and how to properly set this up would be ideal. This way I know for certain I'm testing what you want to include in our project, and that I can verify it works properly. It will also provide me with a reference to share with anyone else who may also encounter issues with setting this up.

.vscode/settings.json
```diff
@@ -12,5 +12,10 @@
     "joypad",
     "input",
     "forms"
-  ]
+  ],
+  "Lua.runtime.version": "LuaJIT",
+  "Lua.workspace.library": [
+    "${3rd}/love2d/library"
+  ],
+  "Lua.workspace.checkThirdParty": false
```
Fellshadow, line 20: These three settings are for the sumneko.lua extension, which you used to have in the recommendations but changed to trixnz.vscode-lua without removing these extension settings. Assuming the person *does* install sumneko's Lua extension there's still some issues with these settings as follows:
- After doing a couple of checks, i've determined that bizhawk runs lua5.1, so the runtime version setting here is incorrect
- Not sure exactly what "Lua.workspace.library" does to be honest, but from what i gathered from a quick search it allows the extension to load external libraries. Not sure what the library you specified here is but i'm pretty sure it's not being used in the tracker code. Not sure if this library would be an extra thing we'd need to install or not / not sure about if this setting is at all needed to be honest

The "Lua.workspace.checkThirdParty" setting is fine (if using sumneko's Lua extension) since as you mentioned this blocks the little "Do you need to configure your work environment as 'LÖVE'?" popup from vscode

Fellshadow (changes requested):
Hey, apologies on this sitting in here so long but to give my two cents here after looking into these things myself a little there's a couple issues here, so i'd be reluctant to merge this in as-is (and I think Zac's wanting to research / try out more of it before he feels comfortable with it). Though IMO these additions would be fine if the issues I've pointed out are resolved.

Agree on the recommended extensions (since we do have a .editorconfig file may as well use it now that I understand that it's for the extension lol). I agree on your reasoning for picking trixnz.vscode-lua over sumneko.lua with it being the one supported by github actions, if we were to get this implemented in the future.

Recently i gave both lua extensions (sumneko.lua and trixnz.vscode-lua) a go installed together and it's working pretty well (I believe Zac has also been trying both, we've already done some cleanup of tracker code via the linting), though tbf I should really try one without the other to see what's being made redundant / if it'd be more useful having both instead of just one.

.vscode/settings.json
davidhouweling: thanks for the review @Fellshadow. I've updated the lua version specified, and i've also removed the workspace library stuff (it was added by default when the popup showed and i clicked apply to workspace.

Fellshadow: ah ok that makes sense then, still useful either way i suppose.
I'm happy with these changes now, though gonna merge to staging instead of main since our workflow now is to push to bugfixes/staging depending on the type of changes and then ultimately pushing things from staging (which regularly gets updated from bugfixes) -> main when gearing up for a new release (which there'll be one soon). Not that it makes a whole lot of difference for these changes, just more for consistency

PR #220, opened 2022-08-27
Bugs: Last Damage inaccuracy and Stat reveal during evo

UTDZac:
This fixes two recent issues. One is seen when you begin a battle and the info stored about that last battle doesnt get cleared out. The other bug is that the tracker reveals information about an evolved pokemon while learning a move, giving you information you shouldnt have yet; you dont have access to the summary screen.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@ Settings.ini
 *.tdat
 *.gba
 *.log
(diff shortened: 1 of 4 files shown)


UTDZac (commented):
Reviewed files.

PR #226, opened 2022-09-04
Make evolution / next learned move stand out more, add stone evos

### LINKED ISSUE #98: Make the next learned move and evo text easier to read
The text in parenthesis for evolution and next learned move can be hard to see in a quick scan due to it not really standing out from all of the text around it.

I think ideally this text could be set to a different color, to set it apart from the rest of the information and provide a bit more visual grouping.

Screen Shot 2022-07-12 at 11 32 18 PM

Fellshadow:
This achieves a couple of things:
- Adjusts the evolution method and level of next learned move colours to make it stand out more and reflect when player's pokemon is close to / able to evolve/learn a new move
  - Closes #98 
  - Currently uses negative text when not close/possible, positive text when close/possible but this could be tweaked to just positive text when close/possible if wanted
  - Achieved by redrawing the text for evo method / next move level with the adjusted colour over the existing text. To be honest, the formulae I used to calculate the spacing/offset for these was found through trial and error. I don't know *why* they work but from my testing they do work with the variable length of the overall line of text due to level numbers
- Slightly reworks the heals in bag updates to also get the amount of evolution stones the player has in each bag
  - This was done without any additional memory reads, as evolution stones also reside in the items pocket.
  - This rework allows for easy extension into also counting the amount of status heals in the bag if we want to do that at any point
- From the tracked evolution stones in bag, the tracker will inform the player that they have a stone that can evolve their current mon

Example images and video of power-levelling through an evo line / entire moveset below:
Untitled

https://user-images.githubusercontent.com/106463662/188328612-9c82f06e-0e78-49d3-9499-c18a559a0c3b.mp4

### DIFF
--- a/ironmon_tracker/Program.lua
+++ b/ironmon_tracker/Program.lua
@@ -119,9 +119,9 @@ function Program.update()
 		end
 	end
 
-	-- Only update "Heals in Bag", "PC Heals", and "Badge Data" info every 3 seconds (3 seconds * 60 frames/sec)
+	-- Only update "Heals in Bag", Evolution Stones, "PC Heals", and "Badge Data" info every 3 seconds (3 seconds * 60 frames/sec)
 	if Program.Frames.three_sec_update == 0 then
-		Program.updateBagHealingItems()
+		Program.updateBagItems()
 		Program.updatePCHeals()
 		Program.updateBadgesObtained()
 	end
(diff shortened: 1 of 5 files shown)


UTDZac (approved):
Awesome work. This definitely looked more complex than I would have though, especially with the stone addition. The working solution here is perfect, and flexible if we want to handle other things.

PR #227, opened 2022-09-05
New dev/debug tool: Add Items

Fellshadow:
Adds a new dev/debug tool that lets you add items to your bag.
Idea for this came from the script thisisatest provided a while back in the discord for adding rare candies.

Usage:
- Load the tool *after* loading the tracker, which will open up a form for the tool and a popup for usage info
- Select the item from a dropdown, and specify quantity in the text box (which i've limited to 2-digit unsigned integers) and click "Add" to add the item
  - For HMs and Key items, you can technically specify a quantity greater than 1 but the tool will set it to 1. I think it still works either way but might as well adjust this just in case
- Item gets added to the bottom of the relevant bag pocket, so that existing items don't get overwritten (unless you have a full bag)
  - Note, if you have the bag open when you click "Add" you will need to close and reopen the bag to actually see the item added

Untitled

Notes:
- Most key items will probably not work, as they require certain flags to be set in the game in addition to having the item
  - On this note, may be interesting to set it up to auto-set those flags later if wanted (i already know how to set the flags for mystic/aurora ticket in FRLG for example) but i decided to not bother for now
- I didn't see anything game-breaking when i've tried this out, but sometimes things may get weird (if the item isn't known by the game then it'll pop up as ?????? for example). I'd highly recommend making a savestate before using the tool, as a restore point just in the off-chance that something *does* go wrong (or you want to revert).
  - It's probably fine for regular items since you can toss, it's particularly key items you may want to be careful about adding

Fun Fact: Apparently Acro/Mach bike are programmed in FRLG to work just like the regular bicycle, so adding/using them works.

### DIFF
--- a/ironmon_tracker/Debug/AddItems.lua
+++ b/ironmon_tracker/Debug/AddItems.lua
@@ -1,0 +1,142 @@
+AddItems = {}
+
+function AddItems.createAddItemForm()
+    local y = 15
+    local x = 10
+    local itemData = {}
+    table.insert(itemData, Constants.BLANKLINE)
+	for _, item in pairs(MiscData.Items) do
+        if item ~= "unknown" then
+		    table.insert(itemData, item)
+        end
+	end
+
+    local addItemForm = forms.newform(170, 115, "Add Items")
+    forms.setproperty(addItemForm, "MinimizeBox", false)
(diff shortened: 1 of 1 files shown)


UTDZac (approved):
Great tool and addition.

PR #228, opened 2022-09-05
Double battle2

jtigues:
- Complete overhaul of the viewSlots system. Now all 4 pokemon's view slots will be tracked for double battles battle.
- Move effectiveness will always use the Left pokemon from the opposing team for both your pokemon and the opponent's. (Note, the opponent's Left is the Player's Right)
- You can now toggle between all 4 pokemon in a double battle in the order: Your Left -> Their Left -> Your Right -> Their Right -> Repeat
- updateViewSlots moved to Battle.lua since it's a Battle-only function


jtigues:
Too many reverted changes, undoing

PR #230, opened 2022-09-05
Double battle overhaul

jtigues:
- Complete overhaul of the viewSlots system. Now all 4 pokemon's view slots will be tracked for double battles battle.
- Move effectiveness will always use the Left pokemon from the opposing team for both your pokemon and the opponent's. (Note, the opponent's Left is the Player's Right)
- Move/Ability tracking is only done for the pokemon on the Left
- You can now toggle between all 4 pokemon in a double battle in the order: Your Left -> Their Left -> Your Right -> Their Right -> Repeat
- updateViewSlots moved to Battle.lua since it's a Battle-only function


jtigues:
- Moved ViewSlot and isViewingLeft to Battle
- Added ownViewSlot and otherViewSlot back to Tracker as dummies to keep old tracker data from being read in wrong

UTDZac (approved):
This has been a long awaited addition, especially for many Ruby/Sapphire/Emerald runs. Thank you for figuring this one out, looks good.

PR #233, opened 2022-09-09
Pc heal auto decrement bug

jtigues:
- Should fix the PC Heals auto incrementing at the start of battle
- Turns out the first thing that happens in a battle is the save blocks are copied elsewhere, and re-encrypted. So either the memory was being read from the wrong place entirely, or we were trying to read data encrypted with the old key using the new key


UTDZac (approved):
Thanks for this quick fix. Makes sense that the PC heals shouldn't ever decrement during battle.

PR #235, opened 2022-09-09
Next move level highlighting

Fellshadow:
### Changes the next move level highlighting in the following way:
- Doesn't highlight if not close to learning new move (as it used to be)
- When 1 level away from learning a new move, will highlight the level of the next move in a colour determined from the theme colours
- This has a couple limitations, as i'll list below, but should at least hopefully be a decent solution for now if we want to take the time for a better one later on. On the whole it reduces the chances of hard-to-read highlight colours.
- I set it up to only calculate the colour when the tracker screen initialises and when the theme settings change, to minimise redundant calculations

### How the colour is calculated:
- The colour is decided between the available colours in the current tracker theme (except for main background, default text, and header text). It's chosen as whichever colour has the highest contrast ratio with the main background colour, so it should in theory be readable on most, if not all themes (if not then the theme probably has readability issues on the whole lol).
- This contrast ratio is calculated using the method given in the [W3C Web Content Accessibility Guidelines](https://www.w3.org/WAI/WCAG21/Techniques/general/G18.html#tests) which is a specification for web-pages that insists on a sufficient contrast ratio between text (and images of text) and the background of the text (among many other things).

### Limitations
- As it's based off contrast ratios, the colour chosen doesn't quite have as much meaning as the highlight for say the evo text, as it won't necessarily be the "positive text"
- Doesn't really take into consideration the header text colour (I tried to toy around a little but didn't get anything that worked consistently well across themes), so the chosen colour won't necessarily be distinct from the header text. Though what this means is that the "worst-case" scenario goes from "distinct from header text *but* unreadable" to "not incredibly distinct from header text *but* is readable" so still an improvement imo, and this limitation only really affects themes consisting of mostly very similar colours


UTDZac (approved):
Thanks for fixing this up. Colors continue to be an annoying thing to work with given the unpredictable ways that end-users end up customizing their Tracker theme.

PR #236, opened 2022-09-09
Merging Bugfixes into Staging [Sept 9th]

UTDZac:
Summary of fixes:
- Finally fixed the auto PC heal count decrementing twice for no reason during battle
- Hiding old save data for heals in bag and badges obtained
- _(Unsure why all of ninjafriend's commits double-up'd here)_


UTDZac (commented):
Re-reviewed the code.

PR #239, opened 2022-09-11
Ability info updates

jtigues:
- bringing all mentions of status conditions to use the same nomenclature & capitalization
- fixing minor spacing issues
- adding more info to Guts, Effect Spore, Soundproof


UTDZac (approved):
Looks good. Thanks for these updates.

PR #244, opened 2022-09-18
Merging Bugfixes into Staging [Sept 18th]

UTDZac:
Summary of fixes:
- Beat Up added as a typeless move for calculations
- Updated startup screen messages about Tracked data to be more clear
- Fixed an issue where the initial attempts counter would be incorrect depending on the filename of the ROM


UTDZac (commented):
Reviewed the changes.

PR #245, opened 2022-09-18
Next release 6.3.1

UTDZac:
This isn't a required release push, as nothing in 6.3.0 has broken on release that requires hotfixing. Mostly doing this to test the compatibility of the new Version Update Checker that was implemented in 6.3.0. With this patch release of 6.3.1, users should **not** be notified of this patch, but they can still find it in the Lua Console or on the release page per normal.

This release includes:
- New option to show a repel icon indicating remaining duration of the repel effect
- More clear and informative messages about tracked data being loaded
- A few minor bug fixes


ironmon_tracker/Program.lua
```diff
@@ -1,5 +1,7 @@
 Program = {
 	currentScreen = 1,
+	inStartMenu = false,
+	startMenuCallback = 0,
```
Fellshadow, line 4: Oh just noticed i forgot to remove this lol

was trying out reading the callback function in the start menu at one point but it didn't work nicely so removed it but accidentally left this variable in

Not anything breaking since it's not doing anything but this `startMenuCallback` can be safely deleted somepoint

Fellshadow (approved):
All good to me, there was one variable I forgot to remove with the repel stuff so I've added a commit to remove that

PR #246, opened 2022-09-20
Double battle overhaul

jtigues:
Major changes
- Fully supported Move and Ability Tracking added for Double Battles
- Ability Tracking support added for Ability changing scenarios (Role Play, Trace, Transform, Skill Swap)
- Enemy moves are now tracked when you transform into an Enemy (only on your next turn at the earliest, when you view your moves)
- Added ability tracking for Lightningrod, Run Away, and a missed case for Insomnia/Vital Spirit (Yawn)

Under the hood
- Added an improved battle action monitoring framework 
- Added a number of GameSettings variables, mostly for the action monitoring, and the aforementioned ability checking
- Transform status and Abilities are now tracked 
- moved "readBattleValues" to Battle.lua
- updated Symbols.lua in Debug to write to a file instead of print
- OneDrive issue fix in inifile


jtigues:
Just mentioning, I'm going to be looking later today at a solution for move effectiveness 'targeting'. The current changes default to the 'Left' pokemon, which is not the greatest for Double Battles.

UTDZac (commented):
I'm still in the process of reviewing this one, notably the `Battle.lua` file which is quite large. Just leaving a note that I'm doing so, and haven't forgotten about this PR.

ironmon_tracker/Battle.lua
```diff
@@ -451,6 +525,14 @@ end
 function Battle.endCurrentBattle()
 	if not Battle.inBattle then return end
 
+	--Most of the time, Run Away message is present only after the battle ends
+	Battle.battleMsg = Memory.readdword(GameSettings.gBattlescriptCurrInstr)
+	if Battle.battleMsg == GameSettings.BattleScript_RanAwayUsingMonAbility then
+		local battleMon = Battle.BattleAbilities[0][Battle.Combatants[Battle.IndexMap[0]]]
+		local abilityOwner = Tracker.getPokemon(battleMon.abilityOwner.slot,battleMon.abilityOwner.isOwn)
+		Tracker.TrackAbility(abilityOwner.pokemonID, battleMon.ability)
+	end
```
UTDZac, line 534: For what purpose is this check required? Prior to this PR, a player's own pokemon's ability would be track automatically as it is used in battle. So when a battle begins and their pokemon is sent out, if it has any ability at all, that ability is automatically tracked.

This check seems unnecessary, but I'm not sure if other code changes require it.

ironmon_tracker/Battle.lua
```diff
@@ -507,4 +602,118 @@ function Battle.changeOpposingPokemonView(isLeft)
...
+	local teamIndex = Utils.inlineIf(isOwn,0,1)
+	Battle.BattleAbilities[teamIndex][slot].abilityOwner.isOwn = isOwn
+	Battle.BattleAbilities[teamIndex][slot].abilityOwner.slot = slot
+	Battle.BattleAbilities[teamIndex][slot].ability = Battle.BattleAbilities[teamIndex][slot].originalAbility
+
+	Battle.BattleAbilities[teamIndex][slot].transformData.isOwn = isOwn
+	Battle.BattleAbilities[teamIndex][slot].transformData.slot = slot
+end
+
+function Battle.handleTransformedMons()
```
UTDZac, line 681: Nit-picky request here, but could this function be renamed to describe what it's doing? I'm not entirely sure what "handling a transformed mon" means under the context of a battle. Is the intent of this function: `trackTransformedMoves()` ?

ironmon_tracker/Battle.lua
```diff
@@ -436,6 +473,43 @@ function Battle.beginNewBattle()
...
+				},
+				["originalAbility"] = ability,
+				["ability"] = ability,
+				transformData = {
+					isOwn = false,
+					slot = i,
+				},
+			}
+		end
+	end
```
UTDZac, line 512: Is this something can be all be put into a separate function, to help keep the "begin battle" function better organized and clean to read. I am also curious about the purpose behind what is going on here. I understand it reads populate battle abilities, but im not sure what that means or what thats for. Can chat about this later if you'd like

ironmon_tracker/Battle.lua
```diff
@@ -284,131 +361,89 @@ function Battle.checkEnemyEncounter(opposingPokemon)
...
 		Battle.Synchronize.battlerTarget = -1
 	end
-
-	--Levitate doesn't get a message, so it gets checked independently
-	if enemyAbility == 26 and Battle.attacker % 2 == 0 then
-		local levitateCheck = Memory.readbyte(GameSettings.gBattleCommunication + 0x6)
-		if levitateCheck == 4 then
-		-- Requires checking gBattleCommunication for the B_MSG_GROUND_MISS flag (4)
-			return true
+	local indexesToTrack = {}
```
UTDZac, line 373: what are "indexes" that need to be tracked? Is this "battle index for the pokemon combatants?"

jtigues: Yes, the return value will be an array of values 0 to 3 inclusive, representing the index of the combatant in the battle (0 amd 2 are allies, 1 and 3 are the enemies).
The ability tracker previously returned true or false to indicate whether the opponent's pokemon's ability should be tracked. Now, it determines the specific pokemon whose ability needs to be tracked. In cases where one ability directly triggers another (i.e. Intimidate vs Clear Body), this allows both pokemon's abilities to be tracked from the one message.

ironmon_tracker/Battle.lua
jtigues: This check is for tracking the ability Run Away. While a prompt appears on the screen for the ability, the battle ends as far as the tracker is concerned a few frames after the message appears, so most of the time it is not caught. So the check for the message needs to occur in the end battle script itself.
This for the case where you skill swap/transform into and enemy mon and then Run. If it was your own ability it would have already been tracked by the logic you mention, which should still be in place in the main battle update function.

ironmon_tracker/Battle.lua
jtigues: Can update the name and yes, this function is two-part
1) it determines if an allied pokemon is selecting a move to use
2) if it is, and we are transformed into an Enemy pokemon, we track all of that enemy's moves

ironmon_tracker/Battle.lua
jtigues: Yes. BattleAbilites is the object which records changes to Abilities or identity (think Transform) over the course of a battle, this is just initializing it with both trainers' pokemon's starting details. So instead of logging whatever ability a pokemon has for them, the program will now know which pokemon that ability came from, and track it for the appropriate pokemon, even if they are no longer one of the 2/4 pokemon in the battle

jtigues:
Oops, tried to rename the branch and apparently that causes github to delete it

UTDZac (approved):
Changes look great. Thank you.

PR #248, opened 2022-10-04
Iconsets: Author credits and periodic icon change

Fellshadow:
Small update:
- Adds credits for who added each icon set
   - Sources for each iconset in our tracker are actual game sprites/portraits so don't know if we need to explicitly credit those anywhere, not sure where there'd be room to do so (could maybe be in README and/or one of the wiki pages if we wanted them?)
- Adds an interval at which the icon randomly changes, in addition to being able to click to cycle them
   - Interval is currently set to effectively every 180 frames, after accounting for the current screen redraw interval (30 frames)
   - Could maybe remove the click functionality if wanted, kept it in if you wanna spam-click to view a bunch in a row lol
   
This screen is now looking a little cramped tbh lol tried to space it out best I could
Don't think we're planning to add new settings here soon anyways (and can deal with the space at that time)

One suggestion I have, if we ever make any new add-ons, would be to have the tracker add-ons be a separate menu screen on the tracker options (either accessed from the setup menu or the navigation menu)

image

### DIFF
--- a/ironmon_tracker/Options.lua
+++ b/ironmon_tracker/Options.lua
@@ -71,7 +71,7 @@ Options.IconSetMap = {
 		extension = ".png",
 		yOffset = 4,
 		adjustQuestionMark = false,
-		author = "Fellshadow"
+		author = "Fellshadow",
 	}
 }
 
--- a/ironmon_tracker/Options.lua
+++ b/ironmon_tracker/Options.lua
@@ -47,6 +47,7 @@ Options.IconSetMap = {
 		extension = ".gif",
 		yOffset = 0,
 		adjustQuestionMark = true, -- Whether to adjust the question mark icons in routeInfo screen
+		author = "Besteon",
 	},
 	["2"] = {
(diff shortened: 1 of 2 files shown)


UTDZac (commented):
This is a great addition. I have two suggestions:

- Can the amount of time before the next Pokemon is automatically shown be extended from 3 seconds to 5 seconds? This gives the user more time to look at the image and consider if they like the particular icon set, or not.
- Can the timer to change Pokemon be reset each time the user clicks an arrow button to cycle through the icon sets? Currently, if I spam click a button, the amount of time before the Pokemon changes is shortened significantly, so much so that I have a hard time comparing a single Pokemon viewed over different icon sets. I believe simply resetting the timer max on-click should solve this.

Fellshadow:
changes made, also made it reset on changing the pokemon icon by clicking it too (so any of the three existing iconset buttons) so you have time to view the one you clicked and got it changed to

UTDZac (approved):

PR #249, opened 2022-10-04
Repel overflow fix

jtigues:
- Variables like the various step counters are stored in an array of words in the saveBlock. For reasons that are yet unclear, sometimes the upper byte is populated with garbage data when using the intro skip patch. This data in the upper byte never clears even when the repel runs out, causing the Repel display to think that there are not only always steps remaining on a repel, but also that it is a ridiculously high number, causing the display to overflow its bounds.
- TL;DR: we're reading the whole word when the data we need is entirely in the first byte.
- Tested for all combinations of patched/unpatched FireRed, both the intro skip and the new hidden item patch.


UTDZac (approved):
Good catch here, and makes sense why this was breaking.

PR #250, opened 2022-10-05
Custom Startup Pokemon Icon

UTDZac:
This is a new feature to allow the user to set the Pokemon icon that is shown on the startup screen. They can choose to show their favorite Pokemon every time game resets, or they can have it based off of the # of attempts, or be completely random.

image


Fellshadow (approved):
Nice addition

PR #251, opened 2022-10-05
Merging Bugfixes into Staging [Oct 4th]

UTDZac:
Summary of fixes:
- Repel usage icon should no longer stay "full"
- Auto PC heal decrement bug actually fixed
- Attack moves without power no longer show incorrect type-effectiveness


UTDZac (commented):
Re-reviewed the changes, looks good.

PR #253, opened 2022-10-06
Ability tracking overhaul

jtigues:
- Better handling of the delay for moves
- Issue was, the sequence of the game code is as follows:

1) Start Pokmeon 1's turn
2) If pokemon is confused, play the confusion message
3a) Use move
3b) Check if move is actually canceled (hurt in confusion, fully paralyzed, truant etc.)
3c) If move is canceled, play relevant script
3d) If we make it here, use the move

If our 30 frame read happened to be in the 1-2 frames immediately before 3b, all the flags would be the default values, aka it would look like the pokemon was using their move successfully. So a skip was added in for move logging; now moves will be updated on the next update after a change in action is detected


UTDZac (approved):
The code looks good to me.

Fellshadow (approved):
ye looks good to me too, what i tested in 1.0 worked fine

PR #254, opened 2022-10-07
Return/Frustration now calculated if close to max

UTDZac:
Small update here to have Return & Frustration show a calculated power value if it's close to being maxed. In such cases, it will only reveal if the power is calculated to be 100 or greater.

The logic is available to reveal the power regardless of friendship/happiness value, but that is generally regarded as hidden information and should not be revealed. This PR is more of a small quality of life update to let you know it's been maxed out.


Fellshadow (approved):
looks good to me

PR #258, opened 2022-10-09
Feature: Clicking on number of Attempts now allows user to edit the counter

UTDZac:
The number of attempts shown on the Startup Screen is now a clickable button. When clicked, it prompts the user to enter a new attempt number. This way users can set an attempt count without having to go into their Tracker folder and update a text file.

This PR also does a bit of cleanup for anything that references a filename, a file postfix, or a file extension.

image


Fellshadow (approved):
seems good to me

PR #260, opened 2022-10-10
Feature: Random Ball Picker

### LINKED ISSUE #256: Feature Request: Option to have randomized ball selection after loading next rom
Nothing fancy, just add an option to enable the function, with the function being as follows:
```
function Main.RandomNextBall
	if Options["Choose Next Ball"] then 
		if seedRolled then
			goto seeded
		else
			math.randomseed(os.time)
			seedRolled = true
		end
		
		::seeded::
		local choice = math.random(1,3)
		if choice == 1
			then ball = "Left"
		elif choice == 2
			then ball = "Middle"
		else
			ball = "Right"
		end
		
		print "For this run, you will be using the " .. ball .. " Ball! All hail " .. ball .. " gang!"
	end
end
```

the ::seeded:: bit is just because lua lacking a "continue" function (at least, as of 5.2), and, obviously, the text is free for modification. 

opinions, thoughts?

UTDZac:
Closes #256 

The random ball picker only appears if the player doesn't have any Pokémon and when the player's character is in "the lab". When it is first shown, it will display a randomly chosen ball. The user can click the re-roll dice button to re-randomize the selection.

This can be disabled through the settings under Gameplay settings: "Show random ball picker"

image

image


Fellshadow (approved):
now people can blame the tracker if they lose the lab fight ;)

PR #263, opened 2022-10-11
Fix for Double Battles created from 2 Singles

jtigues:
- Indexes are different for Double Battles started from two Single Battle trainers seeing you at the same time


UTDZac (approved):
Thanks for the quick fix!

PR #266, opened 2022-10-12
GameSettings cleanup

Fellshadow:
A big ol' cleanup of the big pile of mess that is GameSettings.lua
The GitHub changes comparison below may be hard to read, might be easier viewing the new resulting file directly [here](https://github.com/Fellshadow/Ironmon-Tracker/blob/fellshadow/gamesettings-cleanup/ironmon_tracker/GameSettings.lua)

This should hopefully be a lot simpler and easier to navigate / add new addresses to, but let me know your thoughts.
I've done some brief testing on all currently supported games (i.e. checking it doesn't crash the tracker on load, and playing a tiny bit) and it seems to work fine but definitely want to double-check and do more thorough tests just in case i broke something lol

In the new file, addresses are grouped as such:
- EWRAM (0x02) addresses, which were the same between all versions of each game.
- IWRAM (0x02) addresses, which were the same between all english versions and all non-english versions of a game but different between english and non-english versions
- ROM (0x08) addresses, which were not always the same between versions of a game
- BattleScript addresses used for ability tracking (also 0x08 ROM addresses), which are separate for cleaner groupings

I've left a comment in each section detailing the formatting of the addresses, but here's a brief summary:
- EWRAM: Split by `GameSettings.game` index (i.e. RS/Emerald/FRLG)
- IWRAM: Split first by `GameSettings.game index`, then further split between english / non-english (added in a `GameSettings.language` to help with this alongside loading the localisation data)
- ROM & Ability Tracking: Split first by a locally defined `gameIndex` (which is essentially a split by `GameSettings.versioncolor`), then by a locally defined `versionIndex` for each game

The process for adding a new address would simply be navigating to the relevant section for the address type, checking how it is formatted, and adding the new address in the right format to the local addresses table.


UTDZac (commented):
#1 `GameSettings.setGameInfo(gamecode)` has the potential to fail; when it does, many of GameSettings properties will be nil. As such `GameSettings.setGameVersion()` and all other setup functions should be skipped, otherwise nil errors will occur after "game not supported" dialog prompt appears

#2 Super minor, but can you move `RouteData.setupRouteInfo(GameSettings.game)` outside of `setGameInfo` function and into `GameSettings.initialize`, near the bottom of it. Basically helps organize with the thought that "after game stuff is setup, lets establish route data"

UTDZac (approved):
Incredible work with cleaning this up, thank you!

PR #268, opened 2022-10-15
Last seen level of encountered Pokemon now tracked

UTDZac:
This adds support for tracking and displaying the last level seen for an encountered Pokemon. This helps match move data tracked from previous encounters with the newly encountered, likely higher level Pokemon. To make space for displaying this feature, the old "HP: ?/?" useless text has been removed when viewing the enemy Pokemon.

When encountering a Pokemon for the first time, a blank line is shown:

image

Future encounters will show the level of the previously encountered Pokemon:

image

### DIFF
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -505,6 +505,9 @@ end
 function Battle.endCurrentBattle()
 	if not Battle.inBattle then return end
 
+	-- Only record Last Level Seen after the battle, so the info shown doesn't get overwritten by current level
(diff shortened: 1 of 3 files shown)


UTDZac:
A few notes here:
- This is a change to the Tracker Data object. It should be compatible and upgrade safe, as it doesn't change existing data. It adds a new data property to be tracked. Code that uses this property will check if it's `nil` just in-case.
- Noting the last seen level of a Pokémon doesn't care if it was a wild Pokémon or one that belonged to a Trainer. This is intentional and by design.

Fellshadow (approved):

PR #272, opened 2022-10-17
Battle timing improvement

jtigues:
- More exact battle start/end timings
- Symbols debugging tool overhauled to output in the new GameSettings format
- Moved Battle Delay to Batle.Lua since Program wasn't even using it


ironmon_tracker/Battle.lua
```diff
@@ -464,8 +463,14 @@ end
...
-	Program.Frames.battleDataDelay = 60
-
+	-- Delay reads and display while animations start
+	if Battle.battleStartDelay == -1 then
+		Battle.battleStartDelay = Utils.inlineIf(Battle.isWildEncounter, 100, 40)
+	end
+	if Battle.battleStartDelay > 0 then
+		Battle.battleStartDelay = Battle.battleStartDelay - 10
+		return
+	end
```
UTDZac, line 473: Two things here:

- It feels a bit weird to me that we use `battleStartDelay` as both a boolean "its either -1 or its not" and as a value that goes above and below "0". Unsure if it's better to split this up into two variables to be safe. For example, someone might implement this later with `battleStartDelay = 19` which will later evaluate to -1 and have unintended functionality.
- These two if-statements can likely be connected via an ifelse. It's a little inefficient to set the value, then immediate deduct from that value, the condition is guaranteed true. Just a minor thing concerning the readability with the flow of logic.

UTDZac (commented):
Reviewed and left some notes. It looks like the logic for determining when a battle begins and when it ends has changed a bit, and relies on some other fuctions/messages/reads from the game. I suppose this will be more accurate (hurray!) but I haven't tested them myself, just looked through the code tonight.

ironmon_tracker/Battle.lua
jtigues: Sure, can make a battleAnimationEnded flag to handle the -1 state and combine the if statements

UTDZac (approved):
Debug changes look good to me.

PR #273, opened 2022-10-18
Inner focus tracking

jtigues:
- Added Inner focus script and set it up for tracking


UTDZac (approved):
Nice clean fix, thanks for doing this

PR #275, opened 2022-10-29
Feature: Text shadow option and preset thumbnail previews

UTDZac:
This PR was originally meant to add an additional option to the Theme config that allows the users to turn off the text shadows. Several themes look a bit better if the shadows for text are disabled. This adds an extra option bit to the existing theme config strings (from export/import), and should be compatible for both previous gen3 trackers and the current NDS trackers.

Since the theme customization page was a bit crowded, I moved this boolean options to a "new page." The page ended up a bit empty with only a couple of options, so I may have outdid myself here by adding a fancy thumbnail preview for each of the Theme Presets. These previews are drawn in real time, they are not simply an image capture.

An additional feature to add to this in the future may be a "Save new" button to allow the user to save their current theme into the ThemePresets file. I have not implemented this yet, partly because this was a good bit of work already and partly because I want to save custom themes from users into a different text file that won't get overridden by future tracker updates.

image


Fellshadow (approved):
Neat addition, the preset preview is cool!

PR #278, opened 2022-10-31
Next Release: 6.5.0

UTDZac:
What's new in this release:
- New feature: Enemy Pokémon now display a "last level seen" to help remember what moves that Pokémon might have
- New feature: A built-in Theme Manager to help better manage your collection of custom themes
- A Theme option for removing drop shadows from text on the Tracker, allowing for more crisp looking Themes
- The ability Inner Focus is now automatically tracked if it's triggered
- The move Snatch is now properly tracked when it successfully "snatches" a move
- The room names for the E4 locations in Ruby/Sapphire/Emerald now display properly
- When opening Tracker pop-up windows, Bizhawk sound no longer stutters loudly
- Clarified if a Pokémon was already chosen when clicking the icon on the Startup Screen to change it

I believe I captured all of the changes between 6.4.2 and 6.5.0 and listed them above.


Fellshadow (approved):
Noticed the RSE e4 route data was accidentally reverted so added it back, which is also probably the only thing i'd add to the changelog in the release e.g.
- Added missing E4 rooms to RSE route data

UTDZac:
Good catch, I had completely forgotten we added this.

PR #281, opened 2022-11-03
Synchronize Bugfix

jtigues:
- Fixed issue where Synchronize was not logging if prevented from inflicting a status that was received by the opponent's ability


Fellshadow (approved):
nice clean fix

PR #282, opened 2022-11-03
First turn healing item logging bug

jtigues:
- Didn't consider that allied healing items don't actually take an action
- Now considering the enemy's first turn as well as the ally's to get around it


Fellshadow (approved):
another nice clean fix

PR #283, opened 2022-11-04
Backup last seed before loading the next one

AndreVolpi:
This PR intends to save the last seed played when the user requests a new one, so they can go back to it (for log review or when reset was unintended).

### DIFF
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -437,6 +437,11 @@ function SaveCurrentRom(filename)
 		currentrom:close()
 		historyrom = historyrom:seek("end")
 		historyrom:close()
+
+		olderromfilename = string.format("seed-%s %s", Main.ReadAttemptsCounter() - 5, filename)
+		if Main.FileExists(olderromfilename) then
+			os.remove(olderromfilename)
+		end
 	end
 end
 
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -422,10 +422,10 @@ end
 
 function Main.SaveCurrentRom(filename)
 	if Main.FileExists(filename) then
-		Main.CopyFile(filename, string.format("seed-%s %s", Main.ReadAttemptsCounter(), filename))
-		Main.CopyFile(string.format("%s.log", filename), string.format("seed-%s %s.log", Main.ReadAttemptsCounter(), filename))
+		Main.CopyFile(filename, string.format("seed-%s %s", Main.currentSeed, filename))
+		Main.CopyFile(string.format("%s.log", filename), string.format("seed-%s %s.log", Main.currentSeed, filename))
 
-		olderromfilename = string.format("seed-%s %s", Main.ReadAttemptsCounter() - 5, filename)
+		olderromfilename = string.format("seed-%s %s", Main.currentSeed - 5, filename)
 		if Main.FileExists(olderromfilename) then
 			os.remove(olderromfilename)
 			os.remove(string.format("%s.log", olderromfilename))
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -428,6 +428,7 @@ function SaveCurrentRom(filename)
 		olderromfilename = string.format("seed-%s %s", Main.ReadAttemptsCounter() - 5, filename)
 		if Main.FileExists(olderromfilename) then
 			os.remove(olderromfilename)
+			os.remove(string.format("%s.log", olderromfilename))
 		end
 	end
 end
--- a/Ironmon-Tracker.lua
+++ b/Ironmon-Tracker.lua
@@ -420,7 +420,7 @@ function Main.GenerateNextRom()
 	}
 end
 
-function SaveCurrentRom(filename)
+function Main.SaveCurrentRom(filename)
 	if Main.FileExists(filename) then
 		Main.CopyFile(filename, string.format("seed-%s %s", Main.ReadAttemptsCounter(), filename))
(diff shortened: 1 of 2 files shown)


Fellshadow:
Working at the moment so can't take a proper look at this until later (unles Zac looks it over before i do), just noting that i changed the base branch from main -> staging since main is usually kept in state of current release while staging is where we add new features to in preparation for future releases (bugfixes is the branch for small hotfixes). 

Doesn't make a huge difference right now since we just had a release so main/staging are up to date but in future there may be a case where staging has quite a few things updated from main

Ironmon-Tracker.lua
```diff
@@ -418,6 +420,29 @@ function Main.GenerateNextRom()
 	}
 end
 
+function Main.SaveCurrentRom(filename)
+	if Main.FileExists(filename) then
+		Main.CopyFile(filename, string.format("prev-seed %s", filename))
+		Main.CopyFile(string.format("%s.log", filename), string.format("prev-seed %s.log", filename))
+	end
+end
```
UTDZac, line 428: Could you move the check to see the "file I wish to copy" exists or not inside of the function that performs the copy? This way any other devs who may want to use the `CopyFile()` function don't have to implicitly understand they need to first check to verify the file exists.

Ironmon-Tracker.lua
```diff
@@ -418,6 +420,29 @@ function Main.GenerateNextRom()
...
+	local original = io.open(filename, "rb")
+	local copy = io.open(newfilename, "wb")
+
+	while true do
+		local originalBlock = original:read(2^13)
+		if not originalBlock then 
+		  break
+		end
+		copy:write(originalBlock)
+	end
```
UTDZac, line 440: Small request here for readability, but can you move the condition for the loop to the loops definition itself. For example:

> local nextBlock = original:read(2^13)
> while nextBlock ~= nil do
> copy:write(originalBlock)
> nextBlock = original:read(2^13)
> end

I believe this has the same functionality, just a bit cleaner. Worth double-checking it still works.

Ironmon-Tracker.lua
```diff
@@ -418,6 +420,29 @@ function Main.GenerateNextRom()
 	}
 end
 
+function Main.SaveCurrentRom(filename)
+	if Main.FileExists(filename) then
+		Main.CopyFile(filename, string.format("prev-seed %s", filename))
+		Main.CopyFile(string.format("%s.log", filename), string.format("prev-seed %s.log", filename))
```
UTDZac, line 426: Two small requests here:

1) Can we change format of the previous-file by replacing the "AutoRandomized" portion of the file name with a "PreviousAttempt" label? Should be doable with a simple string `gsub()`.
   - For example, my original ROM is named "FireRed AutoRandomized.gba" and the resulting previous file name is "FireRed PreviousAttempt.gba".
2) Can you add the "PreviousAttempt" text to `Constants.Files.PostFixes`, so all of these file labels can be kept in one place?

Am interested in discussing this further if you have a preference for one method of labeling versus the other.

UTDZac (commented):
Thank you for this small and simple addition. I agree this should be a default feature of the Tracker, given I find myself accidentally starting a new game only then to realize I wanted to see the log info or play my previous game attempt again.

I left a few suggestions on some minor improvements to function, readability, and organization for our project. Let me know if you have any questions.

Ironmon-Tracker.lua
AndreVolpi: [e8672f6](https://github.com/besteon/Ironmon-Tracker/pull/283/commits/e8672f63862017bae44eaeee54cb9b15e809ae64)

Ironmon-Tracker.lua
AndreVolpi: [f38fe80](https://github.com/besteon/Ironmon-Tracker/pull/283/commits/f38fe801fbb91bea7c63fb475d5c29c4dca80a35)

Ironmon-Tracker.lua
AndreVolpi: [3b95fac](https://github.com/besteon/Ironmon-Tracker/pull/283/commits/3b95fac823ebcce95e9f0545fec5dc55c4a35b5e)

Ironmon-Tracker.lua
Fellshadow: also to add to this, may want to perhaps add safety nil checks on `original` and `copy` after their `io.open` calls to confirm they're non-nil before doing anything with them / attempting to close the files

Ironmon-Tracker.lua
```diff
@@ -418,6 +420,28 @@ function Main.GenerateNextRom()
...
+		local nextBlock = original:read(2^13)
+		while nextBlock ~= nil do
+			copy:write(nextBlock)
+			nextBlock = original:read(2^13)
+		end
+
+		original:close()
+		copy:close()
+	end
+end
```
UTDZac, line 443: I want to second Fellshadow's concern here that the two files that are being opened should check for `nil` similar to how all other `io.open` calls are handled. While it's a minor change overall, simply having any nil conflict often causes the tracker script to crash, which in turn many end-users don't often understand what happens or know how to fix it.

I'd recommend dropping the `Main.FileExists(filename)` and add the checks after opening the files:

> if original == nil or copy == nil then
> print(string.format("Error: Unable to create a backup copy of %s."), filename)
> return
> end

Ironmon-Tracker.lua
AndreVolpi: I don't see the reason to check for nil when opening (in theory there's no harm in copying to/from empty files) but I added logic to make it safe to copy (And append to)

UTDZac (approved):
Looks great. Thank you for working with us to on getting this included. I'm excited to have backups of my runs now :)

PR #286, opened 2022-11-07
Merging Bugfixes into Staging: Various ability bug fixes and Battle tracking

UTDZac:
Summary of fixes:
- Fixed bug where Synchronize wouldn't get tracked if it failed to apply status
- Fixed bug with using item on first turn of battle prevents enemy move from being tracked
- Prevent auto-saving tracked data before a game official begins: user hasn't picked their starter yet
- Fixed bug with Damp ability loop where its value could get out-of-bounds


UTDZac (commented):
Re-reviewed the code changes. Looks good.

PR #287, opened 2022-11-08
Feature: Automatic Self-Updater

UTDZac:
This is a new feature that expands upon the "Check for Version Updates" to allow the Tracker to fully and seamlessly update itself (HURRAY!!). The was a bit tricky to setup, and even trickier to test properly, but here is the current structure:

### Changes
- New files are `Main.lua` and `UpdateScreen.lua`. The first hosts mostly old code, with some new version-update additions, the second contains a new screen to show options for updating.
- Most all code has been moved out of the `Ironmon-Tracker.lua` file into a `Main.lua` file. This is because the file that is loaded into Bizhawk cannot be auto-updated without manually reloading the script or restarting the emulator. All the other files can be reloaded live while the emulator is running. Thus, after an update completes, there needs to always be a function exposed from the primary Tracker script to reload all other files.
- The Tracker version has intentionally been set to "1.0.0" to help with testing. Prior to an official release to `main` branch, this will be updated to the next major version: "7.0.0"
- On the Navigation Menu screen for Tracker settings, there is now a 5th button for "Check for Updates" which is always available. This is helpful if you accidentally clicked to skip the current update release, or if a new update came out in the middle of the day after the Tracker's once-a-day version check has occurred. 
   - This button changes to "New Update Available" if the Tracker has already determined that an update is available. This is typically only noticeable under the scenario where the user saw the Version-Update screen originally but chose "Remind me tomorrow" or "Skip this update"

### Feature Explanation
- The Version-Update screen will appear under the same conditions that the old pop-up notifications would appear: typically only once per day and only if an new release is known to exist
- **Update automatically:** This executes a series of batch commands that download, extract, and replacing existing Tracker files with the new ones. Afterwards, it reports the status of the update procedure (See below).
- **Remind me tomorrow:** Informs the Tracker that you don't want to deal with updating right now, hiding this screen until the following day.
- **Skip this update:** Hides the Version-Update screen permanently for this current newest release. You won't be notified of this screen again until a new major/minor update is released in the future.
- **View release notes:** Simply opens a browser window to the latest release on Github, which is typically where release notes are.

### Current Feature Concerns
- While I hope most of this is functional as-is, its quite challenging to test that this will correctly download the latest release, since that won't actually occur until the release update *after* the release that includes this feature. Let me know if you have any questions about how to go about testing this. Most of my testing was setting internal variables to different states so that I could see what things looked like, as clicking the auto-update button would always replace all my code files :(
- This uses batch commands to perform many actions related to downloading, extracting, and replacing Tracker files. While these worked on my own machine and environment, it's possible this may not work for all users' computers. This is challenging to test and ensure that all edge-cases are captured and handled properly.

### The Version-Update Screen
image

### Status Updates after clicking "Update automatically"
image


ironmon_tracker/screens/UpdateScreen.lua
```diff
@@ -0,0 +1,309 @@
...
+		string.format('rmdir "%s\\ironmon_tracker\\Debug" /s /q', folderName),
+		string.format('del "%s\\.editorconfig" /q', folderName),
+		string.format('del "%s\\.gitattributes" /q', folderName),
+		string.format('del "%s\\.gitignore" /q', folderName),
+		string.format('del "%s\\README.md" /q', folderName),
+		string.format('xcopy "%s" /s /y /q', folderName),
+		string.format('rmdir "%s" /s /q', folderName),
+		'echo Version update completed successfully.',
+		'timeout /t 3',
+	}
```
Fellshadow, line 163: I know you mentioned it'd be nice for better error info if something went wrong, here's a suggestion:

Pause the commands when an error occurs. When you're chaining commands like this, if you want to run a command if any of the previous commands failed you can do so like this:
```cmd
(command 1 && command 2 && command 3 && ...) || <command to run if error>
```
So in the lua script could implement it as:
```lua
local batchCommands = {
    '(echo Downloading the latest Ironmon Tracker version.',
    ... --(the other commands)
    'timeout /t 3) || pause',
}
```
Note that the opening bracket goes before the first command and the closing bracket after the last command, since they get joined later. The `||` specifies "run the following command only if the previous command failed", which in this instance the "previous command" is actually the entire batch of commands. This pauses the command output leaving the window open with the error displayed for someone to screenshot and post when asking for help on the issue (instead of it just instantly closing without any indication of what went wrong). 

For example, here I changed the `TAR_URL` to download a tar from a different branch than main, meaning `archiveName` and `folderName` were not correct:
> image

Weirdly though it seems to error there when attempting to copy the files, not on extracting? lol 

If you want to instead just output the errors to a log similarly to the randomiser error log the syntax is very similar:
```cmd
(command 1 && command 2 && command 3 && ...) 2>outputfile.txt
```
though what i noticed this way is that the table for the curl output download gets sent to the output file instead of the command prompt, but that doesn't really matter for the average user

ironmon_tracker/screens/UpdateScreen.lua
```diff
@@ -0,0 +1,309 @@
...
+	local screenToShow = Utils.inlineIf(Program.isValidMapLocation(), Program.Screens.TRACKER, Program.Screens.STARTUP)
+	Program.changeScreenView(screenToShow)
+end
+
+function UpdateScreen.openReleaseNotesWindow()
+	-- The first parameter is the title of the window, the second is the url
+	if Main.OS == "Windows" then
+		os.execute(string.format('start "" "%s"', Constants.Release.DOWNLOAD_URL))
+	else
+		os.execute(string.format('open "" "%s"', Constants.Release.DOWNLOAD_URL))
```
Fellshadow, line 204: this won't work on linux, as `os.execute` itself just straight up fails on bizhawk linux annoyingly from what TheChucklingAtom and BurninTheBlues have said in the discord thread

Fellshadow (commented):
I think it's fine enough to add in, seems to work in the brief testing i did, but left a couple of comments that you should consider whether to implement or not before merging this in

ironmon_tracker/screens/UpdateScreen.lua
UTDZac: Gotcha. I didn't think it would work, but added it in anyway in the event a future Bizhawk version allows this support. I don't have a good way to open a browser window on Linux, nor a Linux system setup to test options.

ironmon_tracker/screens/UpdateScreen.lua
UTDZac: Good notes here, this is something I definitely want to include. I went with option 1 for now, just having it pause if there was an error.

ironmon_tracker/screens/UpdateScreen.lua
```diff
@@ -0,0 +1,309 @@
...
+
+function UpdateScreen.performAutoUpdate()
+	UpdateScreen.currentState = UpdateScreen.States.IN_PROGRESS
+	Program.redraw(true)
+
+	-- Disable Bizhawk sound while the update is in process
+	local wasSoundOn = client.GetSoundOn()
+	client.SetSoundOn(false)
+
+	gui.clearImageCache() -- Unsure if this is needed
```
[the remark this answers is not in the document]
UTDZac: This is what I assumed would happen. I also commonly run into the "Bizhawk is using this files and can't replace them" problem. I'm glad this function fixes that. I haven't fully tested it in all cases, but it seems to do the trick and allow for all of the files to be replaced.

ironmon_tracker/screens/UpdateScreen.lua
```diff
@@ -0,0 +1,309 @@
...
+		string.format('rmdir "%s" /s /q', folderName),
+		'echo Version update completed successfully.',
+		'timeout /t 3',
+	}
+
+	local combined_cmd = ""
+	for _, cmd in ipairs(batchCommands) do
+		combined_cmd = combined_cmd .. cmd .. ' && '
+	end
+	combined_cmd = combined_cmd:sub(1, -5) -- Remove trailing " && "
```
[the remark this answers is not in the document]
UTDZac: Noted. Thanks for the tip.

Fellshadow (approved):

PR #289, opened 2022-11-09
Enhancement: Exposing the Accuracy and Evasion stat changes

### LINKED ISSUE #288: Track Accuracy and Evasion changes in battle
When in battle, it is important for the player and any viewers to know if a Pokémon's accuracy or evasion has been changed due to things like Double Team, Sand Attack, X Accuracy.. etc. Currently this is not being tracked in anyway but it really seems like it would fit in with the toolset of the tracker to do so.

UTDZac:
resolves #288 

The Acc/Eva stats will only be displayed if all of the following are true:
1. At least one of the two stats has changed
2. You are in battle (implied)

Also snuck in a small nil check for something lua-check was yelling about

image


Fellshadow (approved):
Looks good to me, doubt this would cause any conflicts with the other PR as they don't really overlap in file changes so merging this in now

PR #292, opened 2022-11-10
Merging Bugfixes into Staging: Better version comparison and no ssl curls

UTDZac:
Summary of fixes:
- This adds a `--ssl-no-revoke` parameter to the `curl` calls for version checking and archive downloading. In some rare cases, the curl would fail if this param is not present.
- Adding more logical version check comparison. Instead of simply a "not equals" to determine if a new version is out, this will now properly check to see if that version is "newer" than the current one.


UTDZac (commented):
Re-reviewed the code.

PR #293, opened 2022-11-10
Quick hotfixes for upcoming 7.0.0 Release

UTDZac:
This adds two small fixes. This is important as one of them might be preventing automatic updating for some users (ssl revoke).

Summary of changes:
- This adds a `--ssl-no-revoke` parameter to the `curl` calls for version checking and archive downloading. In some rare cases, the curl would fail if this param is not present.
- Adding more logical version check comparison. Instead of simply a "not equals" to determine if a new version is out, this will now properly check to see if that version is "newer" than the current one.


Fellshadow (approved):
seems good to me

PR #295, opened 2022-11-12
Merging Bugfixes into Staging: Several nil checks and RSE RouteInfo additions

UTDZac:
Summary of fixes:
- Fixed a critical missing nil check when Quickloading for the very first time
   - no original autorandomized file to copy
- When performing an auto-update, the Tracked Data is now first saved to the TDAT file, then the update occurs
- Added two extra routes names for trainer battles in RSE:
   - Weather Institute 1F
   - Weather Institute 2F
- Cleaning up some unused variables in `Battle.lua`
- Fixed a missing nil check when using ROM Folder Quickload without first restarting Bizhawk

### DIFF
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -555,12 +555,6 @@ function Battle.endCurrentBattle()
 		RightOwn = 2,
 		RightOther = 2,
 	}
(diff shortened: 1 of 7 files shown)


UTDZac (commented):
Re-reviewed code.

PR #296, opened 2022-11-12
Hotfix 7.0.1 - Critical Fix for Quickload

UTDZac:
Summary of fixes:
- Fixed a **critical** missing nil check when Quickloading for the very first time
   - no original autorandomized file to copy
- When performing an auto-update, the Tracked Data is now first saved to the TDAT file, then the update occurs
- Added two extra routes names for trainer battles in RSE:
   - Weather Institute 1F
   - Weather Institute 2F
- Cleaning up some unused variables in `Battle.lua`
- Fixed a missing nil check when using ROM Folder Quickload without first restarting Bizhawk


ironmon_tracker/Main.lua
```diff
@@ -428,17 +432,21 @@ end
...
+		-- The originalFile to copy doesn't exist, simply do nothing and don't copy
 		return false
 	end
 
 	nameOfCopy = nameOfCopy or (filename .. " (Copy)")
 
 	-- If the file exists but the option to overwrite/append was not specified, avoid altering the file
 	if Main.FileExists(nameOfCopy) and not (overwriteOrAppend == "overwrite" or overwriteOrAppend == "append") then
-		print(string.format("Error: Unable to modify file %s, no overwrite/append option specified."), nameOfCopy)
+		print(string.format('Error: Unable to modify file "%s", no overwrite/append option specified.'), nameOfCopy or "N/A")
```
Fellshadow, line 449: The closing bracket for the string.format is in the wrong place, so the second argument is actually going to the print function here

Same issue in lines 433 (deleted) and 462

Fellshadow (changes requested):
Mostly good, but one breaking bug still there

ironmon_tracker/Main.lua
UTDZac: Ah, that explains so much. I had assumed a different problem. Good catch!

Fellshadow (approved):

PR #297, opened 2022-11-12
New Feature: Pedometer, an optional step counter

UTDZac:
This is a commonly requested feature option, to be able to count steps you take during gameplay. Luckily, this stat is already tracked by the game itself.

#### What this feature offers:
- An option under **Gameplay Options** to enable or disable; it's disabled by default
- When enabled, it occupies the same space on the Main Tracker screen as the Badges carousel
   - Disabling the main screen carousel option forces the pedometer to display instead of the badges
- The step counter shows total number of steps taken so far (max of 999,999; not sure what the real cap is)
- Clicking "**Reset**" will temporarily clear this out, starting back over at 0. Click it again to go back to showing the Total steps
- Clicking "**Goal**" will prompt the player to enter a number of steps to use as a goal post or milestone. When the step counter displays a number equal to or greater to the goal number, it will highlight to show that the goal has been reached.
   - When a goal is set, then the button color changes to indicate that a current goal is active

#### Functionality notes:
- None of this information is saved between runs. The total steps are tracked by the game itself, so no need to track them in the Tracker.
- The player will have to manually set the step goal each time they start a new run or load up an old game. This info isn't saved. I don't think it needs to be saved, since the purpose and use-case may change from game-to-game
- The Gameplay Options screen is getting crowded. We're gonna need to come up with a better organization method soonish

image

### DIFF
--- a/ironmon_tracker/Utils.lua
+++ b/ironmon_tracker/Utils.lua
@@ -29,10 +29,15 @@ function Utils.centerTextOffset(text, charSize, width)
 	return (width - (charSize * text:len())) / 2
 end
 
--- Accepts a positive or negative integer and returns a string formatted as "12,345"
-function Utils.formatNumberWithCommas(integer)
-	if integer == nil then return "" end
-	return tostring(math.floor(integer)):reverse():gsub("(%d%d%d)","%1,"):gsub(",(%-?)$","%1"):reverse()
+-- Accepts a number, positive or negative and with/without fractions, and returns a string formatted as "12,345.6789"
+function Utils.formatNumberWithCommas(number)
+	local _, _, minus, int, fraction = tostring(number):find('([-]?)(%d+)([.]?%d*)')
+
+	-- reverse the int-string and append a comma to all blocks of 3 digits
+	int = int:reverse():gsub("(%d%d%d)", "%1,")
+
+  -- reverse the int-string back remove an optional comma and put the optional minus and fractional part back
+  return minus .. int:reverse():gsub("^,", "") .. fraction
 end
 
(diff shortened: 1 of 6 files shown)


ironmon_tracker/Utils.lua
```diff
@@ -29,6 +29,12 @@ function Utils.centerTextOffset(text, charSize, width)
 	return (width - (charSize * text:len())) / 2
 end
 
+-- Accepts a positive or negative integer and returns a string formatted as "12,345"
+function Utils.formatNumberWithCommas(integer)
+	if integer == nil then return "" end
+	return tostring(math.floor(integer)):reverse():gsub("(%d%d%d)","%1,"):gsub(",(%-?)$","%1"):reverse()
```
Fellshadow, line 35: If the intent is to provide specifically an integer, why does it need to be floor rounded? 
The way this is written seems more like "Accepts a number, converts to integer, and returns a string formatted as '12,345'"

Fellshadow (approved):
Looks good to me for merging in, just got one question regarding the Utils function in terms of clarity on what its intended use-cases are

ironmon_tracker/Utils.lua
UTDZac: Good note. I simply borrowed the code solution from somewhere online. I tested it with an input of `1234.56` and removed the `math.floor()` call, seems to work just fine. So I'll update it as such.

UTDZac: There were some issues with getting it to be compatible with fractional numbers, so I found a different solution that accounts for a wider selection of inputs.

PR #298, opened 2022-11-12
Merging Bugfixes into Staging: Low Kick vs Tower-Ghost fix

UTDZac:
Summary of fixes:
- Fixed an issue where encountering an unidentified ghost Pokemon in the Pokemon Tower while having the move Low Kick would cause the Tracker to crash
- Subsequently fixed other potential look ups where trying to get a Pokemon not within the data set.


UTDZac (commented):
Re-reviewed the code.

PR #301, opened 2022-11-13
Battle starting flag

jtigues:
- New Battle.battleStarting flag which is triggered by the GAME_STAT for total encounters incrementing.
- Using this to shut off reads for PC Heals and the repel tracker, since memory is shuffled around before the battle starts.


UTDZac (approved):
Looks great. Thanks for the quick fix.

PR #302, opened 2022-11-13
Merging Bugfixes into Staging: Attempts count when failed and new Battle start flag

UTDZac:
Summary of fixes:
- More accurate Battle start flag for hiding the Repel icon more quickly
- Attempts counter only increments if it successfully finds a new ROM to Quickload


UTDZac (commented):
Re-reviewed code.

PR #304, opened 2022-11-14
Battle abilities crash

jtigues:
- Fixing an issue where in Emerald, Steven's NPC 3 + 3 tag team double battle hard-coded personality values, including one with "0", which resulted in the pokemon not being read into memory since the tracker ignores 0 as an invalid ID
- Now trainer ID is also checked to determine whether a pokemon block is valid
- Toggling will skip over the NPC pokemon. Eventually I would like to display the NPC pokemon still as a third enemy pokemon, but that requires some refactoring which would delay the fix.
- Moves and abilities for NPC pokemon are now tracked on the same basis as enemy pokemon, instead of always like normal allied pokemon. 

Known TODOs:
- gPlayerPartyCount is hardcoded for Emerald, need to get the addresses for other games as well
- want to test in regular double battles and in FRLG before shipping it

### DIFF
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -400,6 +400,16 @@ function Battle.checkAbilitiesToTrack()
 	end
 	local combatantIndexesToTrack = {}
 
+	--Something is not right with the data; happens occasionally for emerald.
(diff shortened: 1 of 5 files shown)


jtigues:
- Tested in FRLG and in regular RSE doubles
- Addresses added for other games

UTDZac (commented):
Looks good, just added some small formatting fixes and an extra comment to explain a weird removal of code

PR #307, opened 2022-11-26
Adding check for 'using a move' action

jtigues:
- Was checking for non-Move actions for the first action of a battle, and then never again after that. Oops.


UTDZac (approved):
Thanks for this bug fix.

PR #314, opened 2022-12-19
Merging Bugfixes into Staging [Dec 19th]

UTDZac:
Summary of bugfixes:
- Added clarity to the move description for the move Trick
- Resolved an issue where using an item as a first action prevented moves from being tracked.


UTDZac (commented):
Re-reviewed the code changes. Short and simple.

PR #316, opened 2022-12-21
Neutral stat marking

### LINKED ISSUE #309: Enhancement: Want to be able to mark a tracked stat as neutral
Currently when marking a tracked stat for a Pokemon, you only have these three options, in order:
- _(leave empty)_
- Positive, indicated by a `+`
- Negative, indicated by a `-`

This enhancement would add a 4th option for **Neutral**, indicated by a `=` or maybe a circle like `o`. Not sure what would work best, or if `=` looks too much like `-`

image

UTDZac:
Closes #309 

This adds a 4th option for marking a stat on the Tracker: neutral or equal. This is helpful for cases where you test a Pokémon's defensive stat and determine its neither low or high, but still want to mark it instead of leaving it as blank or "untested"

The original order of stat markings is unchanged with the 4th option at the end. So if you cycle through the markings for a stat in goes in order:

1. "blank"
2. Positive, indicated by a "+"
3. Negative, indicated by a "-"
4. Neutral, indicated by a "="
5. (Back to 1. "blank")

image

### DIFF
--- a/ironmon_tracker/Constants.lua
+++ b/ironmon_tracker/Constants.lua
@@ -34,6 +34,7 @@ Constants.STAT_STATES = {
 	[0] = { text = " ", textColor = "Default text" },
 	[1] = { text = "+", textColor = "Positive text" },
 	[2] = { text = "--", textColor = "Negative text" },
+	[3] = { text = "=", textColor = "Default text" },
 }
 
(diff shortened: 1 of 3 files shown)


Fellshadow (approved):
Looks good to me

PR #317, opened 2022-12-21
Allowing the Tracker to autoflush saveRAM

UTDZac:
Since Bizhawk doesn't always flush the saveRAM, this is an attempt to make that happen more reliably.

This solution uses a simple hack of checking the `GAME_STATS` for a change in quantity on the stat "game saves". This change occurs the moment the user presses the "Save the game" button, and NOT when the save actually completes. As a result, I added a short **10 second delay** before telling Bizhawk to flush the save. It's not perfect, but its a simple addition that likely helps save some runs (pun intended).

Let me know if you can think of a slightly better way to do this.


Fellshadow (approved):
should hopefully help resolve some saving issues, don't have any thoughts on improvements other than what ninjafriend has mentioned in the discord thread so suppose we'll see what they find there

PR #319, opened 2022-12-22
Auto save flush

jtigues:
- Some minor changes to the new auto flush feature to flush at the end of the save completion instead of an arbitrary amount of time after starting the save
- Now we additionally read in a 60-frame timer that starts when the save completes (this is used to eventually destroy the "Game saved" message when the timer runs out)


UTDZac (approved):
Looks good. Nice find and thanks for this.

PR #321, opened 2022-12-22
Patch 7.1.4

### LINKED ISSUE #141: Allow for cycling backward through types in Hidden Power
Whether it's a right-click, or some other hotkey, being able to cycle backward through typings in Hidden Power would be convenient.

UTDZac:
Closes #141 

This patch release fixes the following issues:
- [MGBA] Added a RANDOMBALL() command for picking a random starter Pokéball
- Setting your Pokémon's Hidden Power type can now be cycled forwards or backwards
- Autoflush SaveRAM is now more accurate
- Fixed an issue where legacy Tracker versions (7.0.3 and lower) weren't being notified of new update
   - This is why this patch version ends in 4 instead of anything else. (x.x.4 is greater than x.x.3)
- Fixed an issue where the asterisk on "Moves *" wasn't highlighting properly
- Fixed an issue where a player wasn't able to set their own Hidden Power
- Fixed an issue where some moves were spoiled if a player caught and released Pokémon too frequently

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ We'd ideally like to support all non-English versions if we can, progress update
 1. **Get a Supported Emulator**:
    - We recommend using the Bizhawk emulator (Windows/Linux only).
 	   - [Download Bizhawk](https://tasvideos.org/BizHawk/ReleaseHistory) (v2.8 or higher)
-		- If you are on windows, make sure to also download and run the [prereq installer](https://github.com/TASEmulators/BizHawk-Prereqs/releases) first.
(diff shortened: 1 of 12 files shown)


UTDZac:
Note, this is reversing a change to "Track all moves when you capture a new Pokemon and add it to your team." As this was, for some people, definitely over the line.

I think one possible solution to this is to trigger on-level up, the track the last four moves the original Pokemon has, working backwards from that level. This way it will ignore any TMs or moves sketched, since it wont be tracking from the current moveset, but rather the original moveset.

Fellshadow (approved):
Seems good to me, not sure what would work in place for the tracking moves other than what you've suggested but yeah better stay on the more cautious end with it imo

PR #323, opened 2022-12-22
Naming cleanup, Sketch bugfix

jtigues:


Fellshadow (approved):
Seems good to me

PR #325, opened 2022-12-23
Hotfix 7.1.6

UTDZac:
This fixes a **CRITICAL bug** that prevents users on **7.1.0** - **7.1.5** from being able to update while the tracker is active; they instead have to close it down and load just the `UpdateOrInstall.lua` file.

This also actually fixes `io.popen` to instead use an alternative method involving `os.execute`. Tested and it works on Windows and MacOS at least. After this, users should be able to use the simple, alternative Quickload method I worked on.

Also snuck in a nice little update-asterisk to show on the tracker startup screen if an update is available.
image

Note: After this this fix goes live, we can delete the 7.1.5 release as that one is short-lived and buggy, and im using it as a placeholder for this 7.1.6 release.

https://github.com/besteon/Ironmon-Tracker/releases/tag/v7.1.5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ quickload/*.gba
 quickload/*.sav
 quickload/*.jar
 quickload/*.rnqs
+ironmon_tracker/osexecute-output.txt
 ironmon_tracker/images/pokemonAnimated/*.gif
 AnimatedPokemon.gif
 RandomizerErrorLog.txt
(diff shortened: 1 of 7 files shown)


Fellshadow:
Additions from bugfixes:
- Fixed a bug with focus punch tracking

Fellshadow (approved):

PR #331, opened 2022-12-30
Release 7.2.0

UTDZac:
New stuff:
- [Full Randomizer Log Viewer](https://user-images.githubusercontent.com/4258818/210026570-860a32a1-a5f8-4b08-8b39-90fe350ba114.png) is now available after you white out from a battle
- Added the ability to opt-in to dev branch updates to beta test new Tracker builds
- Fixed bugs with tracking certain moves, such as Sketch, Mimic, and Focus Punch

Also after this PR, we will split `staging` into becoming `dev` and change `bugfixes` to `beta-test`. I'll do the updates.

Here is new Log Viewer feature: https://github.com/besteon/Ironmon-Tracker/pull/330

image


ironmon_tracker/data/RandomizerLog.lua
```diff
@@ -0,0 +1,2049 @@
...
+			count = count - 20
+		end
+	end
+	if count >= 1 then
+		print(table.concat(moves, "\n"))
+		moves = {}
+	end
+end
+
+function RandomizerLog.setupMappings()
```
Fellshadow, line 479: potential big thing to note with this, these mappings are for the english names so likely won't work on non-english roms (which aren't officially supported on the randomiser but do still often work)

could be fine for now, as long as it's noted, but yeah may need to come up with a more compact/elegant solution that allows substituting the names via the language data we already have. I guess easiest solution could be straight-up replacing these mappings when setting up language data for the language loaded, depending on the load order of scripts?

Fellshadow (approved):
too much to properly comb through of course (at least without delaying this for a day or so lmao) but functionally it works fine in testing and i didn't notice anything that stood out on skim reading, just a couple of quick notes

ironmon_tracker/data/RandomizerLog.lua
UTDZac: Agree. I have some side notes that logviewer really only works for the currently loaded game (cant load a Ruby log in a FireRed game). And also aware that many of the randomizer logs are likely in english only. I'll make a note of this in the release.

Fellshadow:
also adding, in terms of release notes there isn't a whole lot tbf (depending on how much you wanna sell on the log viewer lol), just the log viewer and option to enable dev-branch updates

PR #332, opened 2022-12-30
Hotfix 7.2.1

UTDZac:
This adds some protective checks (not actual fixes) during the automatic update that is executed while the Tracker is open and running. As of 7.2.0 release, I have seen at least two people attempt to update shortly after starting up Bizhawk, only for it to fail. Not only does the update not complete, it also breaks halfway through the process of overwriting and removing files.

### Goal of this PR
To be extra cautious during the parallel update process (while Tracker is running already). I don't want people corrupting their Tracker and stumbling to have to manually redownload the Tracker and lose files or settings.

So far I've witnessed **two potential causes** for this, both of which are included in this hotfix patch. As a note, this hotfix will be applied immediately to anyone attempting to use the Updater (v7.1.6+), as new code gets loaded after it's downloaded but before the update scripts run.

1. The Tracker files are kept inside a folder structure that includse `OneDrive`
   - This check identifies the folder path contains "OneDrive" and aborts the update process before any files are removed or replaced.
2. The Tracker files are kept on a separate harddrive that isn't the primary harddrive `C:\`
   - This is a simple check on the folder path.

To add to this, I was **not** able to reproduce either of this issues on my local machine.

While it's definitely possible that some users with Tracker files on OneDrive or a non-primary harddrive could still update without issue, there doesn't seem to be a good rhyme or reason for when it decides to break. This is really hard to pinpoint or even reproduce.

### After Review
Please feel free to merge this in as soon as you want and create a new release. You can copy the release notes exactly from the old one (v7.2.0), as I want to keep showing that the major update includes the new Log Viewer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,15 +90,15 @@ Emu-specific Lua documentation:
 
 The primary branches of the Ironmon-Tracker repository are as follows:
 
-- **Main**: This is kept in a state of the latest release. We merge into this branch from staging when we are ready to do the final checks and make a new release.
-- **Staging**: This is essentially the "beta" build of the next release, where the majority of contributions merge into.
-- **Bugfixes**: This branch is for any quick/minor bugfixes. It regularly gets updated to match and pull fixes into the staging branch.
+- **Main**: This is kept in a state of the latest release. We merge into this branch from dev when we are ready to do the final checks and make a new release.
+- **Dev**: This is essentially the "staging" build of the next release, where the majority of contributions merge into.
+- **Beta-Test**: This branch is for test builds that Tracker users can opt-in to trying out. It regularly gets updated with new features from the dev branch.
 
-**Make your PRs to either Staging or Bugfixes, whichever is more fitting for the contribution.**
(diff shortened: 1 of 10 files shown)


UTDZac:
Couldn't sleep, snuck in two additional fixes. One small one with icon colors, the other is larger and fixes a bug.

The Randomizer Log Parser failed for games/logs made with 4.4.0 Randomizer. This is because of the terrible method I chose to implement it, assuming the order of the log sectors remained consistent. Turns out, this isn't the case.

I fixed this by rewriting it to work purely on line number references instead of remove lines from a large table. This had a side-effect of greatly increasing the speed at which a log could be parse.

Fellshadow (approved):
Also added a fix where changing the `Toggle View` setting stopped the tracker from loading, as it would try to update the `Constants.OrderedLists.TIPS` table that got removed a while ago

Deleted all references to that table to be safe, doubt we're re-adding the tips at this point since barely anyone read them 🤷

PR #335, opened 2023-01-04
Hotfix 7.2.5

UTDZac:
Fixes critical issue with move type effectiveness failing when a Trainer has more than one Pokémon.


Fellshadow (approved):
prayge for no further issues with this

PR #339, opened 2023-01-13
mGBA premade seeds quickload fix

Fellshadow:
There was a bug in the mGBA premade seed quickload option where it would get stuck always loading the second ROM in the folder

After some debugging, I found that this was because the tracker successfully updated the attempts count file but then failed to read the correct file on reload so would always re-set the value of `Main.currentSeed`
This is because it was using the ROM name returned from `GameSettings.getRomName()` for the attempts file, which in mGBA returns something like `POKÉMON FIRE` instead of the premade ROMs prefix, so would look for `POKÉMON FIRE Attempts.txt` by mistake

The main fix here is to instead only read the currently loaded rom in bizhawk (so this functions as before on bizhawk) and otherwise use the first rom in the `quickloadFiles.romList` for the `romprefix`. I also moved about the code a little to return early if an attempts file path is obtained via the quickload settings file, as that doesn't require the romprefix at all.

Tested on mGBA 10.0.0 and Bizhawk 2.8 on Windows and both quickloads worked fine on both emus, doubt it'd be any different on 2.9 or non-Windows


UTDZac (approved):
Thanks for this bug fix. I knew there was going to be a really weird edge case found eventually with this Quickload change.

PR #342, opened 2023-02-06
Tracker Customization Support

UTDZac:
This feature adds support for code customizations to be added to the Tracker by anyone. By default, the Tracker will not have any custom code files or custom code execution.

To obtain this, developers need to install the **Custom Code add-on** which is a single file. This file allows the developer to hook into specific areas of the Tracker to execute code, similar to an API, but with a bit more control on our end. 

Since this is an add-on, this feature is upgrade safe. This means that upgrading that tracker will not overwrite or directly break the custom code support. Also worth noting that this is a first attempt at offering custom code to be used within the Tracker, and as such it is unlikely that this captures every possible use-case that a developer might want. Future updates to this add-on are possible, but will require developers to make those updates to their own custom code file to obtain them.

#### In short, this feature:
- Is completely optional, and not included in the regular Tracker download
- Exposes "API"-like functions that developers can use to execute code at specific Tracker runtime events
- Is upgrade safe; custom code isn't overwritten with each Tracker update

#### Things to do after this:
- Include the `CustomCodeSettings.lua` and `~Template.lua` add-on code files to the next release, zipped
- Update the Add-on wiki page to explain how to install this
- Create a wiki page detailing how this custom code add-on works, and examples for using it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ Settings.ini
 quickload/*.jar
 quickload/*.rnqs
 ironmon_tracker/osexecute-output.txt
+ironmon_tracker/Custom/*.lua
 ironmon_tracker/images/pokemonAnimated/*.gif
 AnimatedPokemon.gif
 RandomizerErrorLog.txt
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ Settings.ini
 quickload/*.jar
 quickload/*.rnqs
 ironmon_tracker/osexecute-output.txt
-ironmon_tracker/Custom/*.lua
+ironmon_tracker/Custom/*
 ironmon_tracker/images/pokemonAnimated/*.gif
 AnimatedPokemon.gif
 RandomizerErrorLog.txt
--- a/ironmon_tracker/CustomCode.lua
+++ b/ironmon_tracker/CustomCode.lua
@@ -1,0 +1,128 @@
+----------------------------------------------------------------------------------------------
+-- Avoid modifying this file if at all possible, so that customizations remain upgrade safe --
+----------------------------------------------------------------------------------------------
+CustomCode = {
+	-- To enable custom code to be run by the Tracker, set to 'true'; or set to 'false' to disable all custom code
+	enabled = false,
+	-- A list of custom code files to load, each should match the Custom Code template and return an object with functions
+	filenames = {},
+
+	Labels = {
+		filesLoadSuccess = "Custom Code files loaded",
+		filesLoadFailure = "Failed to load",
+	},
+}
+
+function CustomCode.initialize()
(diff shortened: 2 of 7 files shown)


UTDZac:
Going to change this up in a bit to work better for sharing custom code.

UTDZac:
Done. Thanks for waiting. The `CustomCode.lua` file is now included in the actual tracker project. I separated out the parts that devs will change into the add-on specific code files, which are attached to this message.

- [CustomCodeSettings.lua.txt](https://github.com/besteon/Ironmon-Tracker/files/10670785/CustomCodeSettings.lua.txt)
- [CodeExtensionTemplate.lua.txt](https://github.com/besteon/Ironmon-Tracker/files/10670786/CodeExtensionTemplate.lua.txt)

ironmon_tracker/CustomCode.lua
```diff
@@ -0,0 +1,128 @@
...
+-- Executed after a battle ends, and only once per battle
+function CustomCode.afterBattleEnds()
+	CustomCode.execFunctions("afterBattleEnds")
+end
+
+-- Executed before a button is clicked, and only once per button
+-- button: the button object being clicked
+function CustomCode.beforeButtonClicked(button)
+	CustomCode.execFunctions("beforeButtonClicked", button)
+end
```
Fellshadow, line 128: I feel like the wording of this could potentially be slightly confusing, as it's worded as "before a button is clicked" however looking below at `Input.lua` I can see it's actually *after* the user has clicked on the button, but before the button's "onclick" function is carried out from clicking on it. Maybe something like "Executed when a button is clicked, before the button's functionality is carried out"

The wording "once per button" I think is the most confusing here as it implies it will only ever run once for a given button, when in actuality it runs once for that button *per click of the button* as I don't see anything in these changes that stops it running on successive clicks

Realistically it probably makes no difference though, only misinterpretation I can think of is someone somehow thinking it hooks into the button's initial creation and lets them see its initial state and not actually relating to running alongside its onclick functionality at all

Fellshadow (commented):
Looks good to me, mainly just one thing regarding the wording of one of the CustomCode functions

.gitignore
```diff
@@ -8,6 +8,7 @@ Settings.ini
 quickload/*.jar
 quickload/*.rnqs
 ironmon_tracker/osexecute-output.txt
+ironmon_tracker/Custom/*.lua
```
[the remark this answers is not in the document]
UTDZac: Done, as of most recent commit.

ironmon_tracker/CustomCode.lua
UTDZac: Good notes, i'll update the function name and/or description soon.

Fellshadow (approved):
looks good to me

PR #346, opened 2023-02-14
Refactor most files for loading, drawing, and input checks

UTDZac:
Changing up the process needed to add a new screen to the Tracker. In the past, this would require adding pieces to various files so that the screen would be loaded, initialized, drawn, and input check for buttons clicked. This refactor simplifies that process so they only need to be added in a single place: `FileManager.lua`

This refactor automates the process for adding screens and drawing them, as well as adding individual input check methods to each screen. This implementation is outlined below.

Overall, this should make it easier for new additions, especially for any devs of new extensions. Long term I want to normalize the template for each screen to have `Labels` and `Colors` and other things, as well as better ways to create Button objects to make it even easier.

#### New Initialize Template (automatically applied to screen files with an initialize function)
``` lua
for _, luafile in ipairs(FileManager.LuaCode) do
	local luaObject = globalRef[luafile.name or ""]
	if type(luaObject.initialize) == "function" then
		luaObject.initialize()
	end
end
```

#### New Drawing Template (automatically applied to screen files with a drawScreen function)
``` lua
if Program.currentScreen ~= nil and type(Program.currentScreen.drawScreen) == "function" then
	Program.currentScreen.drawScreen()
end
```

#### New Input Template (automatically applied to screen files with a checkInput function)
``` lua
if Program.currentScreen ~= nil and type(Program.currentScreen.checkInput) == "function" then
	Program.currentScreen.checkInput(xmouse, ymouse)
end
```


UTDZac:
Forgot to leave a comment for the context of this being a PR. I mostly just wanted a second set of eyes to quickly skim and review these changes. Maybe there's something obvious I missed that you might want to search for and see if it was updated properly.

I did some light testing on my end for all three emulators (Biz 2.8, Biz 2.9, MGBA) and things appear to be in working order.

Fellshadow (approved):
seems good to me

PR #347, opened 2023-02-17
7.3.1 Release

UTDZac:
This release adds several bugfixes and additional features (optional) for extensions. Developers can now add a url for users to view their extension details online, as well as offer ways to edit extension options and check for updates. This patch also refactors code related to creating new screens to make it easier for devs to add additional screens.

Summary of changes:
- The Tracker's Extensions menu now includes a link for getting more extensions online
- Custom Extensions now support a URL, a version update check, and a way to configure options. If you're an extension developer, refer to the `CodeExtensionTemplate` to learn more.
- When a friendship evolution is ready, the Tracker now says "READY" instead of "SOON" to clarify it's ready to evolve on next level-up
- Cool new icons for some Navigation menu buttons
- Fixed an issue where cycling through Pokémon in a doubles battle would sometimes skip a Pokémon, preventing you from viewing it entirely
- Fixed an issue where the wrong theme colors were used for the text/shadows on some Tracker screens


ironmon_tracker/screens/SingleExtensionScreen.lua
```diff
@@ -0,0 +1,232 @@
...
+		text = "Lower box text",
+		border = "Lower box border",
+		boxFill = "Lower box background",
+	},
+	column2offsetX = 50,
+	extension = nil,
+	extensionKey = nil,
+}
+
+-- TODO: Might add a "Reload Extension" and a "Remove" option later, uncertain yet
```
Fellshadow, line 24: Just adding thoughts:
- If "Remove" means unloading *and* removing associated files, that'd be tricky because you'd need to also account for any files made by the extension itself in runtime (e.g. settings files or files intended for logging / adding as stream sources) and only partially removing / leaving artifacts is a bit messy imo
   - "Unload" could be good though, if that's not currently done when disabling an extension
- "Reload" option could be good if e.g. the extension comes across some weird rare blip that crashes it, without having to entirely reload the tracker (though of course a recurring crash wouldn't be solved by reloading lol)

Fellshadow (approved):
Seems good to me

ironmon_tracker/screens/SingleExtensionScreen.lua
UTDZac: Good notes on "Remove." Chances are this functionality isn't really needed from the Tracker UI. Given you have to install extensions by adding files to a folder on your computer, it's trivial to remove those extensions by simply removing those files from the folder.

As for "Reload" I incidentally included this as part of the ON/OFF enabling of the extension. When you turn an extension ON from an OFF state, it will reload the extension fresh, as if it were the first time you ever use it. Thus, it will have all of the new updates to that file (if you made any). This decision makes more sense to me then keeping the old loaded script. There was some end-user confusion on trying to update an extension while the Tracker was running, but replacing the files and turning it off and on again didn't actually do anything; so that is now replaced with a "reload."

PR #349, opened 2023-02-19
Parallel/Live update now requires Bizhawk restart

UTDZac:
Due to continued conflicts with updating the Tracker script while many of the files are locked by Bizhawk, resulting in corruption, this is a change to only allow parallel updates immediately after a Bizhawk restart.

This is a somewhat wonky solution due to the way Bizhawk remembers/reloads lua script files during a Quickload. So determining when Bizhawk was "actually restarted" versus when it was just loading up everything again after a new ROM load is tricky.


Fellshadow (approved):
Hopefully this helps most, if not all, of the updating issues people have

PR #352, opened 2023-02-24
Adding option to show an EXP bar

UTDZac:
This adds an option to allow the player to add a cool experience points bar on the Tracker screen. This feature means that exp info for the full player's team is read in and can be displayed. Currently, only the lead Pokemon's exp is shown (as normal), but in a future update i want to have a full party display that includes exp bars for each Pokemon in the party.

This PR also changes the default visual behavior of the info line for Level + Evolution. If the Pokemon does not evolve, then nothing is shown at all, implying no evolve.

image


Fellshadow (approved):
Seems good to me

PR #353, opened 2023-02-24
Add an option to turn off friendship calculations

UTDZac:
This adds an option to the Tracker to allow the player to disable the Tracker from telling them when their Pokemon has enough friendship for certain things, such as a friendship evolution is ready or Return's power is maxed.

There weren't too many places that required changes to implement this. So I think I got them all. Also worth noting the Return/Frustration are behind two options now (intentional). In order for it to show you when you're close, you have to enable both "Calculate variable damage" and "Determine friendship readiness", as its possible some people want friendship readiness for evos but dont want move powers calculated for them.

image

### DIFF
--- a/ironmon_tracker/Constants.lua
+++ b/ironmon_tracker/Constants.lua
@@ -112,6 +112,7 @@ Constants.OrderedLists = {
 		"Show physical special icons",
 		"Show move effectiveness",
 		"Calculate variable damage",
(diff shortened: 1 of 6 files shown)


Fellshadow (approved):
looks good to me, with the space available I can't really think of anything clearer than "Determine friendship readiness" lol

PR #355, opened 2023-02-24
7.3.2 Release

UTDZac:
Release Notes:
- Added an [EXP bar](https://user-images.githubusercontent.com/4258818/221065010-77261f6d-53e2-449e-ae6e-e6f878ebc32a.png) to show how many experience points your Pokémon needs to level up.
   - To turn this enable, enable it through Settings -> Gameplay -> "Show experience points bar"
- When battling trainers, the number of [Pokémon on their team](https://user-images.githubusercontent.com/4258818/221167414-74ad6de1-8e64-46dc-9341-8583786ad391.png) and which have been KO'd is now shown
- Add a Tracker option to turn off revealing information about Pokémon friendship. Settings -> Gameplay -> "Determine friendship readiness"
   - When turned off, friendship evolutions will no longer show "READY" and Return/Frustration values will not be calculated for you.
- The Tracker's self-updater now requires you to restart Bizhawk to update, which should fix many issues encountered while updating.
   - The process is as simple as choosing to update, then restarting Bizhawk, then continue with the update.
   - Note: If at any time the update fails, you can always restart Bizhawk and load the `UpdateOrInstall.lua` file by itself to safely upgrade.


Fellshadow (approved):
Seems good to me, may also want to note in the release notes that return/frustration *also* require the "Calculate Variable Damage" setting, for those that might want to disable that but keep the friend evo information

PR #356, opened 2023-02-26
edit log parse pattern strings to be more flexible

Aeiiry:
This was primarily for myself, as I was having issues with the view "view log" fucntion when using the ironmon randomizer https://github.com/something-smart/ironmon-randomizer and/or settings trainers to all be doubles battles for whatever reason. Shouldn't mess up existing logs, but I haven't tested it extensively outside of old logs from the regular randomizer I had lying around.


UTDZac:
For each pattern you've changed, can you speak to why these changes are necessary and perhaps share examples? When I originally created these patterns, I thoroughly tested them for all the games people typically play for Ironmon.

ironmon_tracker/data/RandomizerLog.lua
```diff
@@ -2,11 +2,12 @@
 RandomizerLog = {}
 
 RandomizerLog.Patterns = {
-	RandomizerVersion = "Randomizer Version:%s*([%d%.]+)%s*$", -- Note: log file line 1 does NOT start with "Rando..."
+	-- Acount for possible -dev or -beta in version string
+	RandomizerVersion = "Randomizer Version:%s*([%d%.]+).*$", -- Note: log file line 1 does NOT start with "Rando..."
```
Aeiiry, line 6: This change is simply to have the pattern match the whole line if there are any characters after the version number, could be including in the capturing group. Likely not a needed change

ironmon_tracker/data/RandomizerLog.lua
```diff
@@ -2,11 +2,12 @@
...
 	RandomizerSeed = "^Random Seed:%s*(%d+)%s*$",
 	RandomizerSettings = "^Settings String:%s*(.+)%s*$",
 	RandomizerGame = "^Randomization of%s*(.+)%s+completed",
-	PokemonName = "([%u%d%.]* ?[%u%d%.'%-♀♂%?]+).-",
+	PokemonName = "(%u%a*%s?[%a%d%.’'%-♀♂ ]*%s?[%a%d%.’'%-♀♂ ]*)",
```
Aeiiry, line 10: The ironmon randomizer does not have fully capitalised pokemon names in the log file. Comparison between the regular and ironmon randomizer: 
image
image

Additionally the pattern was changed to be more generic, matching an uppercase letter, any number of following letters, an optional space, then 1-2 instances of the character set that can be found in pokemon names. Looking at this now I can trim it down a bit so I will make another commit to do so.

ironmon_tracker/data/RandomizerLog.lua
```diff
@@ -106,7 +107,7 @@ function RandomizerLog.formatInput(str)
 	str = str:gsub("’", "'")
 	str = str:gsub("é", Constants.getC("é"))
 	str = str:gsub("%[PK%]%[MN%]", "PKMN")
-	str = str:match("^%s*(.*)%s*$") or str -- remove leading/trailing spaces
+	str = str:match("^%s*(.-)%s*$") or str -- remove leading/trailing spaces
```
Aeiiry, line 110: the .* in the original was capturing any trailing whitespaces as it is a greedy quantifier. The %s* would always match 0 spaces. Changed it to .- to match as few characters as it could before matching any whitespaces with %s*

ironmon_tracker/data/RandomizerLog.lua
```diff
@@ -7,7 +7,7 @@ RandomizerLog.Patterns = {
 	RandomizerSeed = "^Random Seed:%s*(%d+)%s*$",
 	RandomizerSettings = "^Settings String:%s*(.+)%s*$",
 	RandomizerGame = "^Randomization of%s*(.+)%s+completed",
-	PokemonName = "(%u%a*%s?[%a%d%.’'%-♀♂ ]*%s?[%a%d%.’'%-♀♂ ]*)",
+	PokemonName = "(%u%a+[%a%d%.’'%-♀♂%s]*)",
```
Aeiiry, line 10: Shaved the pattern I made down significantly

ironmon_tracker/data/RandomizerLog.lua
Aeiiry: Okay I feel like a real goober but its not specific to the ironmon randomizer, I just tend to tick "lower case pokemon names". Which would break the parsing

ironmon_tracker/data/RandomizerLog.lua
UTDZac: My regex is not very great when it comes to understanding the need for one thing over another. As-is, the current code works great for parsing the game files and showing all of the information it needs to show to the player. Can I ask, what does this change actually do that makes a significant change? This question is partly because I'm simply not very good at regex and don't really understand this impact, and partly because I want to learn more about why you'd use `.-` over `.*`

Can you please share an example of a string that is properly formatted with your change to `.-` that is not properly formatted with the current code `.*`? (Not saying either code is better than the other, I'm just merely wanting to understand how this works and how it's beneficial)

ironmon_tracker/data/RandomizerLog.lua
UTDZac: This is a simply enough change that can catch any future Randomizer updates without needing to be so strict. As such, I think it's a good change and should be included.

ironmon_tracker/data/RandomizerLog.lua
UTDZac: A note on this pattern: I expect to be reworking the parser to accommodate different Randomizer settings and custom hacks to Pokémon names, moves, etc. You can read more about this in #357 

Until then, I'd like to hold off on altering this pattern. This pattern is very important and needs to be tested thoroughly. There are weird circumstances where sometimes the names that you see in your log aren't the same as the ones seen in other logs for other people. For example, sometimes `Nidoran ♂` appears as `Nidoran ?` which is why there is a whole separate set of functions to deal with parsing this information properly. I see that your updated pattern lacks the `%?` and may not properly capture these cases.

I'm willing to make changes here, but only if they have been thoroughly tested and only if you can explain in detail the need for each one, and what they accomplish. With that said, don't feel the need to do this as I expect I'll be wanting to rework this system to be more compatible with a variety of customizations people use.

Aeiiry: I certainly underestimated the complexity a bit here. I might have a go at testing different randomization settings/games today just to make sure things work, if that doesn't work out I'll leave this alone for your expected rework.

ironmon_tracker/data/RandomizerLog.lua
Aeiiry: Absolutely!
The quantifier " * " is 'greedy', so it will match between 0 and unlimited instances of the character before, but match **as many as possible** before checking the next part of the pattern.

In the pattern "^%s\*(.\*)%s\*$" , because " . " matches any single (non-newline afaik) character, and because it will match those characters as many times as possible it will always match potential whitespace at the end of the string. The %s*$ after the capture group won't ever match anything as it has the option to match nothing and the .* before it can match the same thing.

The change I made was to use the " - " quantifier, which is the same as the " * " quantifier but will match as **few** characters as possible, so it will stop matching characters as soon as the next part of the pattern ( in this case %s*$) matches and therefore trailing whitespaces will be matched, but outside of the capture group and not contained in the returned string

Aeiiry: Here is some example code:

image

And the output strings for that code:

image

ironmon_tracker/data/RandomizerLog.lua
UTDZac: If you want to update the PR to exclude this section, that would work for me.

ironmon_tracker/data/RandomizerLog.lua
UTDZac: Thank you for the great examples. I see what you're saying here. I suppose this was just kind of working before because the outer patterns themselves were excluding the leading/trailing spaces.

UTDZac:
If you have time to test and verify this on a FRLG log and an Emerald log, each from Randomizer 4.6.0 and Randomizer 4.4.0, and verify this works for all four cases, that would be great. I recommend checking through a variety of pokemon with weird names, such as Mr. Mime, Farfetch'd, Nidoran(Male), and checking through the elite four teams.

I have limited time myself and won't be able to verify these changes thoroughly.

Aeiiry:
I've started doing testing of the games and randomizers above, will update this PR when I'm done.

Aeiiry (commented):
I ended up needing to add some more code to explicitly handle evolutions, as there wasn't an elegant solution with lua pattern strings and their limited capabilities.

Aeiiry:
All changes including my most recent commit tested without encountering any issues.
Randomizers tested:
PokeRandoZX version 4.6.0 and 4.4.0
[PokeRandoZX ironmon edit version 1.5.0(https://github.com/something-smart/ironmon-randomizer)

Roms tested:
Pokemon Emerald
Pokemon Firered

Methodology:
For each combination of rom and randomizer, and also for "lower case pokemon names" enabled and disabled:
* Set up the quickload settings in the tracker to use the rom and randomizer being tested, set the settings file to be a kaizo ironmon settings file compatible with pokerandozx v4.4
* Randomize the game with the tracker's quickload
* Play through the opening sequence until starter pokemon can be made to faint
* Click "view log"
* Ensure bulbasaur and chimeco show info from the log. If any pokemon wasn't parsed correctly in the log then all pokemon after it will not be parsed correctly either
* Check split evolutions for 2,3 and >3 evolutions. I usually checked wurmple, tyrogue, and eevee
* Check elite 4 teams populated correctly and check the individual pokemon on their teams populated correctly.

Additionally I tested enabling as many other randomizer settings as I could and doing the same tests above (randomized types, moves, etc)

No issues encountered

UTDZac:
Thank you for doing some tests. I noticed your change to the **PokemonName** pattern causes issues with some log files. This change excludes some possibilities. The most important difference I notice is that your pattern excludes the possibility of a `?` character, which does show up in some logs for some randomizers and some games. There may exist other differences.

I've attached an example log: [RSE Kaizo lower names AutoRandomized.gba.log](https://github.com/besteon/Ironmon-Tracker/files/10887492/RSE.Kaizo.lower.names.AutoRandomized.gba.log)

PokemonName Pattern change:
- Old: `PokemonName = "([%u%d%.]* ?[%u%d%.'%-♀♂%?]+).-",`
- New: `PokemonName = "(%u%a+[%a%d%.’'%-♀♂%s]*)",`

Ultimately, your change needs to still include all possible matches the previous pattern included, while also including the adjustments you want to make. The challenge is being careful about assumptions of what the log file looks like, as there exist many other variations of log files that we've stumbled upon that cause issues. Sorry this isn't better documented.

At some point I'll have more time to dig into this more and maybe help find a good solution, but currently out of time to dig into this thoroughly. I've personally not great at patterns, so it often takes me a long time to test all the cases.

PR #358, opened 2023-03-01
Add Issue Templates

Fellshadow:
Adds issue templates for the following:
- Bug reports
- Feature Requests
- Custom Code Extension submissions/ideas

The templates provide a form for the user to fill out, and auto-applies the relevant label to the issue. The default blank issues can still be made if desired

If you want to try out how these look, I have a temporary testing repo [here](https://github.com/Fellshadow/TestingIssueTemplates/issues/new/choose) (I'll most likely delete this repo later on when these templates are applied here and we're happy with them, so I don't mind if you actually submit issues to try it out lol)

Note: These won't actually take effect until this is merged into the main branch, we could directly commit this to main if wanted but there's a couple of recent commits from dev that would also be added

PR #359, opened 2023-03-01
Add Issue Templates

Fellshadow:
See #358 for details, just re-making from a new branch so we can put it directly into main with nothing else


UTDZac (approved):
Looks great. A solid addition to help others report problems or submit fun ideas. Thanks for doing this.

PR #360, opened 2023-03-04
Feature: Pokémon Team Preview

### LINKED ISSUE #348: Feature Request: Full Party Pokemon Tracking
While I do understand this is 100% intended for the Ironmon challenge and that the primary pokemon is really the only one that is important, I would love to see the tracker be able to view or output the information about the entire party.

I use this over other trackers like PokeLink for non-ironmon play, and i think it would be a great feature to have.

My initial thoughts is to output the data that you use for the main pokemon, but into text files instead -- at least in terms of minimal information (dex#, Name/Nickname, level, Current/Max HP, Experience/Experience to next level, status (poison,para,etc), held item, etc.)  At least that way, whoever wants can act on that information and display it on their stream can do so, even if it's just displaying changing text.
In this instance - multiple text files would be better than less -- 01dex#, 01nickname, 01status, - - -etc.

Additional feature would be to add a separate browsersource or window for OBS to capture all the party pokemon, but I know this is probably far outside the scope of what this tracker is meant to do.

UTDZac:
closes #348 

This feature adds an option to display a summary of the player's six Pokémon on their team. To enable this feature, it's under Tracker -> Setup -> **Show Team View**.

The summary boxes for each Pokémon on the team include:
- Its nickname
- Its icon and any status afflicted _(clickable)_
- Its types _(clickable)_
- Its current level
- Its HP and EXP, both as percentage bars stacked on top of each other
- Its held item, if any _(clickable)_
- Its ability _(clickable)_

image


UTDZac:
One thing I didn't account for is showing an egg in the party, which doesn't really come up in FRLG. Wasn't sure how to acquire one in this game and to test it. Right now I believe it just won't display on the party, which I think is fine. You can't really do anything with that information anyway.

jtigues (approved):
All looks good to me. Mentioned the minor point of having a minimum width for the individual box widths to allow for the icon + type icons at a minimum, but with the game screen and tracker staying a constant size, shouldn't ever be an issue

Fellshadow (changes requested):
I had an egg testing file to hand, it skips over to the next pokemon creating a duplicate (this is what `Tracker.getPokemon` does to handle eggs. Not sure of the best way to handle this without changing that functionality, you could check if the pokemon retrieved is already in `TeamViewArea.PartyPokemon` but would need to be cautious of discluding actual non-egg duplicates by mistake (though personality values should be different there)

It doesn't affect FRLG too much as that's post-game *but* you do get eggs in the main game of RSE (the Wynaut in Lavaridge). The day care in RSE is also on route 117 and accepts two mons so very possible to breed in main game too, hence why I marked as "request changes" for now (as this is more of a feature intended for e.g. nuzlockers where they might actually want to grab an egg in those games idk). If you think it's fine to deal with this later then i'm not opposed to merging it in now as it looks fine otherwise

screenshots:
image
image

UTDZac:
Support for showing eggs without revealing information is now live.

image

Fellshadow (approved):
yeah seems fine to me now

PR #362, opened 2023-03-04
View log button in 'extras' screen

Aeiiry:
Adds a "view the log" button to the extras screen, with an additional warning screen to prevent players from accidentally clicking the button or dissuade them if they are tempted.
image
image


UTDZac (approved):
This looks great. Did some quick tests with different quickload setups and this functioned as expected.

PR #364, opened 2023-03-10
Update Battle.lua

jtigues:
- Fixing issue where Stat Stages were not updating properly in double battles


Fellshadow (approved):
Simple fix at least, looks good to me

PR #365, opened 2023-03-12
7.4.1 Patch

UTDZac:
Release Notes:
- Fixed an issue where during double battles the stat changes on enemy Pokémon would sometimes be incorrect
- Fixed an issue where during double battles the move effectiveness would be calculated against an already KO'd Pokémon
- Fixed an issue where, in rare cases, wild encounters would incorrectly be displayed as a trainer battle
- [mGBA] Cleaned up some battle tracker display information


Fellshadow (approved):
looks good to me

PR #366, opened 2023-03-13
7.4.2 Release Patch

UTDZac:
Patch 7.4.1 fixed several issues with doubles battles but introduced a new issue. The target cursor information wasn't providing accurate details for the cases where it was being used. Thus, when Pokémon were KO'd in doubles battles, the new slots/targets were incorrectly calculated for purposes of move effectiveness.

This patch resolves this new issue.

Release Notes to append to 7.4.1:
[7.4.2] Fixed yet another doubles battle related issue where move effectiveness was inaccurate if trainer team had 3 or more Pokémon.


jtigues (approved):
Checked on both true (Tate & Liza) and artificial (Brawly's gym) doubles, fix is working fine.

PR #368, opened 2023-03-13
Post e4 roamer battle bugfix

jtigues:
- Fixed bug where Tracker would enter a battle state after the enemy mon memory was used for Latias/Latios in Emerald


Fellshadow (approved):
thanks for the quick fix!

PR #369, opened 2023-03-18
Feature: Show pre-evolutions in log viewer

Aeiiry:
https://user-images.githubusercontent.com/19582161/226071465-891b9b56-4764-46e6-b5b7-0cf53cfce42b.mp4

This allows users to explore evolution trees via clicking on the pokemon icons.

When there is not enough space on screen for all evolutions, chevrons are drawn below the evolution arrow, 1 for each "evo set". Clicking on the arrow or the chevrons will scroll through the evo sets (see video)

If there is no pre-evo, up to 3 evolutions will be shown, matching current functionality.

Logic exists to scroll through pre-evo sets as well, including chevrons but this is far less common.


ironmon_tracker/Drawing.lua
```diff
@@ -89,32 +89,55 @@ function Drawing.drawRightJustifiedNumber(x, y, number, spacing, color, shadowco
...
+		end
+	elseif direction == "right" then
+		x = x + thickness + 2
+		while i < thickness do
+			gui.drawLine(x + i, y, x + i + width, y + (height / 2), color)
+			gui.drawLine(x + i + width, y + (height / 2), x + i, y + height, color)
+			i = i + 1
+		end
 	end
 end
```
Aeiiry, line 140: This function didn't cover left and right chevrons. While adding left and right I also added the option to pass a specific color instead of "hasColor". Old functionality for up and down chevrons remains.

ironmon_tracker/Drawing.lua
```diff
@@ -89,32 +89,55 @@ function Drawing.drawRightJustifiedNumber(x, y, number, spacing, color, shadowco
...
-
```
Aeiiry, line 117: This function didn't cover left and right chevrons. While adding left and right I also added the option to pass a specific color instead of "hasColor". Old functionality for up and down chevrons remains.

ironmon_tracker/data/DataHelper.lua
```diff
@@ -578,6 +578,17 @@ function DataHelper.buildPokemonLogDisplay(pokemonID)
...
+	data.p.prevos = {}
+	for _, prevoId in ipairs(pokemonLog.PreEvolutions or {}) do
+		local prevoInfo = {
+			id = prevoId,
+			name = PokemonData.Pokemon[prevoId].name,
+		}
+		table.insert(data.p.prevos, prevoInfo)
+	end
+
+
```
Aeiiry, line 591: Tracking pre-evos as part of the buildPokemonLogDisplay data table

ironmon_tracker/data/RandomizerLog.lua
```diff
@@ -205,6 +205,11 @@ function RandomizerLog.parseEvolutions(logLines)
 			evo = RandomizerLog.alternateNidorans(evo)
 			if RandomizerLog.PokemonNameToIdMap[evo] ~= nil then
 				table.insert(RandomizerLog.Data.Pokemon[pokemonId].Evolutions, RandomizerLog.PokemonNameToIdMap[evo])
+				-- Add pre-evolutions to the evolved Pokemon
+				if RandomizerLog.Data.Pokemon[RandomizerLog.PokemonNameToIdMap[evo]].PreEvolutions == nil then
+					RandomizerLog.Data.Pokemon[RandomizerLog.PokemonNameToIdMap[evo]].PreEvolutions = {}
+				end
+				table.insert(RandomizerLog.Data.Pokemon[RandomizerLog.PokemonNameToIdMap[evo]].PreEvolutions, pokemonId)
```
Aeiiry, line 212: Tracking pre-evos as they are read in from the file

ironmon_tracker/Drawing.lua
UTDZac: The following function has changed and affects other areas of code. While you have implemented a workaround, I'd like to see this function written more clearly so it could be understand. The function header should, in my opinion, be

```lua
Drawing.drawChevron(x, y, width, height, thickness, direction, color)
```

Where the final parameter `hasColor` is now just `color`. The inner code will have a default color value: `local color = color or Theme.COLORS["Default text"]`. Then, all other calls to this function need to be updated to fit this new header.

In this way, it remains clear the intent of the function by passing a color and the resulting outcome.

ironmon_tracker/data/RandomizerLog.lua
UTDZac: A small note on code legibility here. Since `RandomizerLog.PokemonNameToIdMap[evo]` is being used in more places and is quite the chunk to read, I'd say it's a good idea to assign it to a local variable prior to the first `if` condition here. In this way, the code can be understood a bit more quickly and clearly.

```lua
evo = RandomizerLog.alternateNidorans(evo)
local evoPokemonId = RandomizerLog.PokemonNameToIdMap[evo]
if evoPokemonId ~= nil then
	table.insert(RandomizerLog.Data.Pokemon[pokemonId].Evolutions, evoPokemonId)
	-- Add pre-evolutions to the evolved Pokemon
	if RandomizerLog.Data.Pokemon[evoPokemonId].PreEvolutions == nil then
		RandomizerLog.Data.Pokemon[evoPokemonId].PreEvolutions = {}
	end
	table.insert(RandomizerLog.Data.Pokemon[evoPokemonId].PreEvolutions, pokemonId)
```

UTDZac (commented):
Thanks for putting together this PR. I think it's great to try out cool features like this, but I'm not sure if this is one I want to include in the Ironmon Tracker as-is. I would love to discuss these sorts of features with you prior to working on them if you intend to submit them as part of the project, which may help avoid any extra work or adjust needed after the fact.

There are a few reasons I'm against adding this feature as it is currently:
1. The main reason is that the majority of Ironmon players do not need to see this information by default. While yes, its cool to see this info, it isn't exactly helpful to know about after a run is over. It's not a question many players ask, because of the way ironmon works. Typically the question is "What would be Pikachu evolve into?" and not "I had a Pikachu, I wish I knew what evolved into it". I think a good solution here is to not show the pre-evolution info unless the player opts into wanting to see it, by click a button or toggle to view that information for the Pokemon they are currently inspecting. Am open to discuss other possibilities.
2. Paired with the above concern, screen space within the Log Viewer at a premium. If pre-evos are shown as a default, this would mean that now the majority of Pokemon need that extra space to show off an evolution line. As a result, the Move list gets squished and can't all be seen at a glance. Now when viewing most Pokemon, you can't see their full moveset on a single page, and it requires clicking through pages. I think this takes away from a primary use-case of the Log Viewer in favor on an addition that isn't of that much use by the majority.
3. It is not intuitive that the Pokemon surrounded by a border is the one you are looking at. To me, it's important that the thing you see first on the page is what's helps you know what Pokemon you are looking at. The name as text and the unlabeled border aren't a strong enough indicator. It is also worth noting that you cannot trust Theme colors to do the work for you, because there exist many themes that have invisible borders (box border and background are the same color).
4. The location of the "how does this evolve into that" information seems to be inconsistent. While it is possible to discern what it means for the variety of Pokemon you can inspect, the template isn't the same from Pokemon to Pokemon. I think the evo information being directly below the evolved Pokemon is the most clear to me, instead of being a mix of sometimes under a Pokemon and sometimes under an arrow. We can work this out and discuss it more in the discord channel.

ironmon_tracker/Drawing.lua
UTDZac, line 1: Two issues here.

1. You are defining `color` as an integer value, therefore you cannot use that color in `Theme[chevronColor]`, because its not a string index. One good tip for working on the Tracker is anytime you make a change to colors on the tracker, be sure to test with the default theme and a light colored 2 part theme (like Beach Getaway) and maybe another one as well. This helps show problems with forgetting a color or the color turning `nil`
2. You can bypass a step of needing to make a new variable `chevronColor` and just reuse `color`. This is a clever little Lua trick to reuse a parameter variable but ensure its not nil, thanks to the `or`.

```lua
color = color or Theme.COLORS["Default text"]
```

Aeiiry: I'm slightly confused by your first point, this statement wouldn't change the chevronColor unless it found a valid result in Theme.COLORS.
```
if Theme.COLORS[chevronColor] then
		chevronColor = Theme.COLORS[chevronColor]
```

if you're referring to `local color = Theme.COLORS["Default text"]` that was removed as part of my original changes

I use the beach getaway theme by default and the current code works for chevrons existing before my change and the new usage in the log viewer.
image

image
image

I didn't originally reuse the `color` param to avoid having it have 3 different possible types (string, bool, or int). I'll change it back to re-use color as the scope of this function is small and changing types isn't a big deal.

Aeiiry: Changed chevronColor back to color https://github.com/besteon/Ironmon-Tracker/pull/369/commits/75c29da099452ae358e31a1979660f8ac327f3f8

ironmon_tracker/Drawing.lua
UTDZac: I wrote this code segment in a hurry, but you shouldn't need to redefine `color` as `local color` since it already exist as a param. Instead, simply use `color = color or ...`

You can see other examples of this trick in a few other methods. Examples:
- `Drawing.drawBackgroundAndMargins()`
- `Utils.calcShadowColor()`

Aeiiry:
Current state of the feature right now. Setting added to enable the feature, disabled by default. 

https://user-images.githubusercontent.com/19582161/226103317-5fbf80f0-810d-4b75-9d67-3234eead5ffc.mov

Video shows the game loading with the setting enabled purely because I was testing the setting's ability to load.

ironmon_tracker/screens/LogOverlay.lua
```diff
@@ -93,6 +96,7 @@ LogOverlay.Windower = {
 		end
 		LogOverlay.refreshTabBar()
 		LogOverlay.refreshInnerButtons()
+		Main.SaveSettings()
```
UTDZac, line 99: The `Main.SaveSettings()` call should not be here, and should be called when the user clicks on the option to change it. Check out how other checkbox options work for an example. See `TimeMachineScreen.Buttons.EnableRestorePoints`

ironmon_tracker/screens/LogOverlay.lua
```diff
@@ -408,7 +412,39 @@ LogOverlay.NavFilters = {
...
+		end,
+		toggleState = Options[LogOverlay.preEvoSetting],
+		toggleColor = "Positive text",
+		onClick = function(self)
+			self.toggleState = not self.toggleState
+			Options.updateSetting(self.text, self.toggleState)
+		end,
+	}
+
+	table.insert(LogOverlay.Buttons, preEvoSettingButton)
```
UTDZac, line 447: You should create this button with all the other buttons outside of the initialize() function. After doing this, add a line of code inside this initialize() function to apply the loaded setting to update the button and its state.

`TimeMachineScreen.Buttons.EnableRestorePoints` is an example of how this is done. Check the `TimeMachineScreen.initialize()` function to see how the setting gets applied after the Tracker finishes setting up and has all the settings loaded from the Settings.ini file.

ironmon_tracker/screens/LogOverlay.lua
Fellshadow: This was something we briefly discussed in the discord when it was being worked on, not sure of the performance impact tbf but i'm not sure how much i agree with saving (which involves making a write to the settings.ini file) every time the setting is toggled on/off. Imagine a scenario where someone maybe accidentally double-clicks, then has to click again to re-toggle. That's 3 file writes there potentially

ironmon_tracker/screens/LogOverlay.lua
```diff
@@ -915,71 +960,203 @@ function LogOverlay.buildPokemonZoomButtons(data)
...
+		-- Add prevos to list
+		for i, prev in ipairs(data.p.prevos) do
+			table.insert(preEvoList, {
+				name = PokemonData.Pokemon[prev.id].name,
+				id = prev.id
+			})
+		end
+	else
+		hasPrevEvo = false
+	end
```
UTDZac, line 981: Just noting a chance to do a little clean-up here if interested. You don't need to do an `local hasPrevEvo = [boolean] or false` since the [boolean] part will either be true, or false if false; so that's implied already. You can also merge the conditions that follow as such (I believe this is the same resulting outcome):

```lua
local hasPrevEvo = Options["Show Pre Evolutions"] and #data.p.prevos > 0
if hasPrevEvo then
	-- Add prevos to list
	for i, prev in ipairs(data.p.prevos) do
		table.insert(preEvoList, {
			name = PokemonData.Pokemon[prev.id].name,
			id = prev.id
		})
	end
end
```

UTDZac (commented):
Overall great improvements. I do want to applaud you for being able to dissect the mess that is the `LogOverlay.lua` code I wrote, which is kind of all over the place and not well organized. It was our first attempt at something large like this that was had scalable UI elements.

The setting (Default:OFF) to show pre-evos resolves my main concern from earlier. Thank you for setting this up.

Two final things I'd like to work on are as follows, using the image below as reference
1. The evo information labels underneath the pokemon icons that explains how to evo. This info should always be under the evolved pokemon. In the sample image below, Slakoth evolves into Zangoose at lv18. As such, the "Lv. 18" info should be directly under Zangoose. This way the evo info remains consistent regardless of if I'm viewing Slakoth, or Zangoose, or (if zangoose had an evo) another evo down the line.
2. I still firmly believe there needs to be something else changed such that when viewing a pokemon page, it is very clear that you are viewing the page for that Pokemon. Using the reference image, the first icon i see (the one ive been used to seeing for 3 months now) is slakoth. However, i am not viewing slakoths page. This is a UI decision that needs to be addressed but i dont have a good solution. The result should be that, at a glance, it should be absolutely obvious and clear which pokemon page im viewing based on the icon that I can easily see, which should be Zangoose.

image

ironmon_tracker/screens/LogOverlay.lua
Aeiiry: Went ahead and implement this, makes it a lot more readable ty

ironmon_tracker/screens/LogOverlay.lua
Aeiiry: Moved per your suggestion in latest batch of commits

Aeiiry:
Current functionality, selection indicator added and evo methods are now duplicated for split evos with the same evo method (wurmple, tyrogue)


https://user-images.githubusercontent.com/19582161/226150616-dd782d24-96e7-4157-bb89-a9d07d60e66d.mov

UTDZac:
Looks great. I'm going to merge this in to `dev` and `beta-test` so I can have a few people try it out. Thank you for being able to make the adjustments and have this feature look awesome. It's pretty fun to click backwards to see how the evo things work out.

PR #370, opened 2023-03-20
Log Parser: Adding support for custom Pokémon names

UTDZac:
For cases where the player has modified their rom to replace Pokemon names and move names with custom names. The log viewer now properly parses *most* of this information and displays it. There are some cases where this can fail, such as duplicate custom names (no unique identifier) or weird symbols.

There are quite a few wonky changes necessary to get the parser compatible with custom named logs, so feel free to just skim through this quickly. Mostly just wanted a second set of eyes if anything looks completely out of place here.

I tested on a custom log from a streamer where many Pokémon names and moves are renamed. I also test that it can properly load up different logs and use all info from that log instead of relying on info from the game rom itself. In the past, the abilities of pokemon and the types of pokemon were just retrieved from the rom, but this fails when wanting to view a previous log from a different rom (future feature)

### DIFF
--- a/ironmon_tracker/data/DataHelper.lua
+++ b/ironmon_tracker/data/DataHelper.lua
@@ -537,7 +537,6 @@ function DataHelper.buildPokemonLogDisplay(pokemonID)
 	data.p = {} -- data about the Pokemon itself
 	data.x = {} -- misc data to display, such as notes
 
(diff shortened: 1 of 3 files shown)


Fellshadow (approved):
seems good to me, nothing stood out from a quick look-over

PR #371, opened 2023-03-21
Add button to view previous log

Aeiiry:
I also fixed the tracker always showing the game over screen when exiting the log overlay, now it should only show the game over screen if an actual game over has occurred before the log overlay was shown.

Video shows me navigating the deoxys in the log viewer just to show that all pokemon are parsed with the recent log viewer changes in dev

https://user-images.githubusercontent.com/19582161/226546464-c2c447da-d9c5-42f2-b27d-fa28287baaf5.mov

### DIFF
--- a/ironmon_tracker/data/RandomizerLog.lua
+++ b/ironmon_tracker/data/RandomizerLog.lua
@@ -1,5 +1,7 @@
 -- A collection of tools for viewing a Randomized Pokémon game log
-RandomizerLog = {}
+RandomizerLog = {
+	currentLog = nil,
+}
 
 RandomizerLog.Patterns = {
 	RandomizerVersion = "Randomizer Version:%s*([%d%.]+).*$", -- Note: log file line 1 does NOT start with "Rando..."
--- a/ironmon_tracker/data/RandomizerLog.lua
+++ b/ironmon_tracker/data/RandomizerLog.lua
@@ -1,6 +1,6 @@
 -- A collection of tools for viewing a Randomized Pokémon game log
 RandomizerLog = {
-	currentLog = nil,
+	loadedLogPath = nil, -- Holds the path of the previously loaded log file. This is used to check if a new file needs to be parsed
 }
 
 RandomizerLog.Patterns = {
--- a/ironmon_tracker/screens/ExtrasScreen.lua
+++ b/ironmon_tracker/screens/ExtrasScreen.lua
@@ -113,6 +113,7 @@ function ExtrasScreen.createLogViewButtons()
 				if RandomizerLog.Data.Settings ~= nil then
 					LogOverlay.parseAndDisplay()
 				else
+					LogOverlay.openLogFilePrompt()
 				end
 			end
 
--- a/ironmon_tracker/screens/ExtrasScreen.lua
+++ b/ironmon_tracker/screens/ExtrasScreen.lua
@@ -9,8 +9,8 @@ ExtrasScreen = {
 		resultAboveAverage = "Above average!",
 		resultDecent = "Decent.",
 		resultUnavailable = "Estimate is unavailable.",
-        viewLogFile = "View log",
-		viewPreviousLogFile= "Previous log",
+		viewLogFile = "View log",
+		viewPreviousLogFile = "Previous log",
 	},
 	Colors = {
 		text = "Lower box text",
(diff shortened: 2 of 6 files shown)


ironmon_tracker/screens/ExtrasScreen.lua
```diff
@@ -9,7 +9,8 @@ ExtrasScreen = {
 		resultAboveAverage = "Above average!",
 		resultDecent = "Decent.",
 		resultUnavailable = "Estimate is unavailable.",
-		viewLogFile = "View the log",
+        viewLogFile = "View log",
```
UTDZac, line 12: Looks like the spacing is using actual spaces here instead of tabs. Can you revert this back to using tabs like the line above and below it.

ironmon_tracker/data/RandomizerLog.lua
```diff
@@ -1,5 +1,7 @@
 -- A collection of tools for viewing a Randomized Pokémon game log
-RandomizerLog = {}
+RandomizerLog = {
+	currentLog = nil,
```
UTDZac, line 3: Can you add a comment to explain what this variable is? At first I thought it held the parsed log information for the currently loaded log file, but then I remembered this info is already kept as part of `RandomizerLog.Data`. If this is meant to be something else, can you add a description here for that.

ironmon_tracker/screens/ExtrasScreen.lua
```diff
@@ -171,5 +231,39 @@ function ExtrasScreen.drawScreen()
 	end
 
 	local ivBtn = ExtrasScreen.Buttons.EstimateIVs
-	Drawing.drawText(topboxX + 4, ivBtn.box[2] + ivBtn.box[4] + 1, ivBtn.ivText, Theme.COLORS[ivBtn.textColor], shadowcolor)
-end
\ No newline at end of file
+	if ivBtn then
+		Drawing.drawText(topboxX + 4, ivBtn.box[2] + ivBtn.box[4] + 1, ivBtn.ivText, Theme.COLORS[ivBtn.textColor],
+		shadowcolor)
+	end
```
UTDZac, line 237: Out of curiosity, why is the if-condition needed here? What changed made this required.

ironmon_tracker/screens/ExtrasScreen.lua
```diff
@@ -171,5 +231,39 @@ function ExtrasScreen.drawScreen()
...
+		rompath = FileManager.prependDir(romname)
+	end
+
+	local logpath = FileManager.getPathIfExists((rompath or "") .. FileManager.Extensions.RANDOMIZER_LOGFILE)
+	if logpath == nil then
+		return false
+	end
+
+	return LogOverlay.parseAndDisplay(logpath)
+end
```
UTDZac, line 269: As this function looks more or less identical to `GameOverScreen.viewLogFile()`, I think it would be best to re-use that function here. Provide a parameter for the function `postfix` to provide the choice to replace `FileManager.PostFixes.AUTORANDOMIZED` and `FileManager.PostFixes.PREVIOUSATTEMPT`. Then update the areas of the code that call the function in question with the appropriate parameter, for example:

`GameOverScreen.viewLogFile(FileManager.PostFixes.AUTORANDOMIZED)`

ironmon_tracker/screens/LogOverlay.lua
```diff
@@ -33,7 +33,8 @@ LogOverlay = {
 	currentEvoSet = 1, -- Ideally move this somewhere else
 	prevEvosPerSet = 1,
 	evosPerSet = 3, -- Ideally move this somewhere else
-	preEvoSetting = "Show Pre Evolutions"
+    preEvoSetting = "Show Pre Evolutions",
```
UTDZac, line 36: Looks like the spacing is using actual spaces here instead of tabs. Can you revert this back to using tabs like the line above and below it.

UTDZac (commented):
A few minor requests, but overall looks good.

I see you took a stab at a better way to build buttons for screens, so they are a bit more dynamic. While most of the code has been using the old static way for a while, I do want to eventually move to using a simpler process for creating button objects and laying them out on screen. That change will likely requiring overhauling most code, yours and mine. Just something to keep in mind moving forward when creating things in a different way.

ironmon_tracker/data/RandomizerLog.lua
Aeiiry: This variables just holds the log path of the previously loaded log, or is nil if no log has been loaded. Added this to distinguish between previous log and log for current run when checking if log data exists to load into the overlay.

Its used in LogOverlay in this block at line 474-477

```	
-- Check for what log we're trying to display, and if it's already been parsed
	if RandomizerLog.loadedLogPath ~= logpath then
		RandomizerLog.Data = {}
		RandomizerLog.loadedLogPath = logpath
	end
```
added a comment to explain this and renamed the variable

ironmon_tracker/screens/ExtrasScreen.lua
Aeiiry: This was previously needed before I moved the creation of the log view buttons to their own function, something about the timing of the onClick functions messed up and crashed when clicking/drawing any of the buttons but the log ones. This is not needed anymore though so I'll remove it

UTDZac (approved):
Thanks for making the changes. Looks great.

PR #373, opened 2023-03-23
7.5.1 Patch

UTDZac:
Release Notes:
- [7.5.1] Fixed an issue with clicking View Log would open a file prompt if using the Premade ROMs Quickload setting.
- [7.5.1] Fixed inaccurate encounter rate percentages for Route 103 in Emerald


jtigues (approved):
Looks good.

PR #375, opened 2023-03-23
7.5.2 Patch

UTDZac:
Release Notes:
- [7.5.2] Fixed bug where log viewer would not update to the current log for generated roms
- [7.5.2] While using Bizhawk 2.9-rc2, form popups like taking notes no longer crash the Tracker
- [7.5.2] The Last Damage taken label during doubles battle is now a bit more clear


Fellshadow (approved):
looks good to me

PR #377, opened 2023-03-27
InfoScreen Header Adjustments

Fellshadow:
- Adds `Drawing.drawHeader`, which functions similarly to `Drawing.drawText` but instead using `Constants.Font.HEADERSIZE` as the default font size (set to 16)
- Changes the InfoScreen headers from size 12 bold font to size 16 regular font, to space the text out better horizontally to be more legible
- As a result, elements on the InfoScreen pages for pokémon and moves had to be adjusted, as the header takes up more horizontal space than before

Example images of how the changed InfoScreens look with the update:
image
image
image

### DIFF
--- a/ironmon_tracker/Constants.lua
+++ b/ironmon_tracker/Constants.lua
@@ -16,6 +16,7 @@ Constants.SCREEN = {
 
 Constants.Font = {
 	SIZE = 9,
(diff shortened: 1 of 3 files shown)


UTDZac (approved):
Looks great. Thanks for making these changes.

PR #378, opened 2023-04-08
7.5.3 Patch

UTDZac:
Summary of changes:
- Better error handling to prevent rare Tracker crashes
- Cleaner UI for screens that use headers
- Fixed a rare extension startup bug

While this is an extremely small patch, it's basically just that: a patch. I'm wanting to get the error handling stuff pushed live before any larger update, like an 8.0.0 version, to make sure it works okay and doesn't confuse users. I also wanted to show off the new header changes. Finally, these needs to be compatible with the new auto-randomized rom patch that Krizz is working on.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ quickload/*.rnqs
 extensions/*
 !extensions/How to use.txt
 !extensions/CodeExtensionTemplate.lua.txt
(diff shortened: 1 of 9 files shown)


Fellshadow (approved):
Seems good to me 👍

PR #379, opened 2023-04-21
Adding Japanese game support

### LINKED ISSUE #62: Support for non-English Pokémon versions

UTDZac:
This adds compatibility support to allow the Tracker to play Japanese FireRed ROMs, similar to how it can play other non-English games. This is only now possible because of the support for Japanese character rendering in Bizhawk 2.9. As for mGBA, it technically displays these text characters fine, but the formatting is still messy because there is no way to properly determine the length of a UTF16 string in this context.

The translations were imported from this [Github repo](https://github.com/roytam1/rtoss/tree/master/PokemonHackSourceCode/PokemonMemHack).

I had to do some hacking in `GameSettings` to allow for Japanese ROM addresses, mostly cause EWRAM and IWRAM were different. I want to say that 99% of the addresses are correct, but it's possible some of them may not work.

image


UTDZac:
Forgot to note, the changes to `Battle.lua` were mostly to account for `nil` addresses or undefined ability addresses.

ironmon_tracker/GameSettings.lua
```diff
@@ -401,6 +411,10 @@ function GameSettings.setGameVersion(gameversion)
 				versionName = "Pokemon Feuerrote",
 				versionIndex = 6,
 			},
+			[0x00630000] = { -- Japanese
+				versionName = "Pokemon FireRed J",
+				versionIndex = 7,
+			},
```
Fellshadow, line 417: Could be worth mentioning the `v1.0` here? (either directly in `versionName` or as a comment, just so the note is there for anyone else who looks at this)

Fellshadow (approved):
Seems fine enough to me, hopefully addresses work fine 👀

ironmon_tracker/GameSettings.lua
UTDZac: Part of why i left this out is my lack of research to know what rom i was using and what roms are possible. I *think* the rom im using is 1.0. and I *think* its the only one that exist. but im not entirely sure. I dont have a different version that i found that i could test.

Fellshadow: hm yeah, that's fair i suppose

PR #382, opened 2023-05-20
Update Battle.lua

jtigues:
- Trace double bugfix
- TLDR assumption about the battlerTarget in doubles was incorrect, turns out its usually the other enemy mon.


Fellshadow (approved):
Looks good to me, thanks for the fix!

PR #385, opened 2023-05-30
Fix Levitate ability revealing under some conditions when it shouldn't

jtigues:
- Fixes an issue where the attacking pokemon was updated from memory about a frame before the levitate message was cleared, so very rarely the attacking mon would have its ability tracked at the end of levitate proccing.


UTDZac:
Looks good, thanks for the quick fix

PR #386, opened 2023-06-01
Log Search Screen - Add Spacebar button

Aeiiry:
Simple change to add a spacebar button to the log search screen keyboard. 
Allows for better searching for things such as "dragon rage" or "fire punch"


ironmon_tracker/Constants.lua
```diff
@@ -494,7 +494,8 @@ Constants.CharWidths = {
     ["9"] = 4,
     ["="] = 4,
     [","] = 2,
-    ["-"] = 2,
+	["-"] = 2,
+	["_"] = 5,
```
UTDZac, line 498: Looks like you replaced tabs with spaces. Can you revert this back to make sure it's using tab indents only?

UTDZac:
If you end up adding a clear button `[X]` next to the back arrow, then I think you can also do away with the clearing of search text when the filter dropdown box changes. I know originally I had you clear this out, but if the clear button is there then its really easy for the user to clear it out themselves.

I think a more common use case is to start typing "THUNDER" then realize they are on the "Pokémon Name" filter, changing it to "Levelup Move" which clears out all that text they wrote in.

So yeah, I you had it right the first time, just needed a way to single button clear out.

ironmon_tracker/Constants.lua
Aeiiry: The indentation in that file was/is using spaces for some values in that table. I guess I accidentally converted the indentation for the "-" entry to a tab via my IDE
image

UTDZac: Oh that's strange. I wonder how long it's been like that.

I'll make this update on dev later then, don't worry about this. Thanks for noticing it.

Aeiiry:
image
I've added a clear button to the right of backspace.
Currently the backspace image is misaligned vertically but that's a minor issue.
Additionally I moved the labels for "Sort by:" and "Filter" down 1px as they were slightly too high.

UTDZac:
All of this looks great. Thanks for adding in the extra buttons

Issue #376, opened 2023-03-26
[Feature Request]: Timer, screenshot, and nickname ideas


Tipzilla:
### Tell us about your request

I've been getting into more and more Kaizo attempts myself and while doing so, have thought of a few ideas for the tracker:

- Timer: I've noticed a few streamers like to show a timer for how long the current attempt is going, but it requires them doing manually toggle it all themselves. I thought it would be helpful if the tracker showed a timer (reading from the gamer file time) for each attempt. Instead of it taking up space on the tracker, maybe it could be a pop-up window? 

- Screenshot button: It would be nice if players could easily take a screenshot of the tracker screen so they can share across social media. Currently I just use the Windows Snipping Tool, but it would be handy if you could set a save location and have the tracker sort it all for you. The button could be a little camera icon next to the settings gear?

- Toggle between Pokémon name and nickname: Most of us know the default names of all the Pokémon, so being able to toggle between that name and what you nickname it would be nice. This could either be done through the tracker menus, or simply by just clicking the Pokémon's name for it to switch between the two.

- Nickname suggestion database: This was an idea I've brought up in the Discord. Sometimes I struggle to think of a cool nickname for my Pokémon, I thought it would be cool if the tracker recommended you a nickname when you select your mon. As a community we could have some sort of database that we can add nicknames to that the tracker can read from. Until that database gets populated, it could use randomized names to start with.

Let me know what you think!

Tipzilla:
Another possible idea, purely for a little bit of convenience. Held TM's show an asterisks indicating they are a gym TM, if you click it it will tell you which gym leader it's from. If you pick one up in the overworld the tracker could do a pop-up to say "You would x gym leader's TM!".

UTDZac:
Thank you for all these great ideas for Tracker features. And sorry it's taken this long to reply to your post.

I wanted to share that I've added the "Timer" feature as part of an upcoming release (8.0.0). You can read more about it here:

- https://github.com/besteon/Ironmon-Tracker/pull/391

As for the other items, it looks like taking a screenshot is a supported function of Bizhawk. It captures the entire game screen + tracker screen. I'll look into where the best place is to add this. The nickname ideas are a bit trickier to include, mostly due to the way text/letters are received from the game to the tracker.

TheRealTaintedWolf:
I was able to mock the up the nickname a bit. Should we always uppercase it or something for formatting? It looks like you are already pulling in the nickname from the pokemon's data in Program.lua

[...]

PR #391, opened 2023-07-11
On-Screen Gameplay Timer

UTDZac:
This PR adds a simple timer to the game screen. This is a feature request from Issue #376 .

Please don't merge this PR in until the Resources PR has been merged in; it uses the required resource code changes.

#### Feature Notes
- The Timer can be turned on or off in the Extras Settings screen
- The Timer always appears in the bottom-right of the game screen itself, overlayed on top of the game itself
   - The time is formatted as HH:MM:SS, but omits the HH if play time is under 60 minutes
   - In a future patch, I will likely offer more options for relocating the timer
- The Timer can be paused/unpaused by clicking on it.
   - A help tip is displayed when the timer is first enabled to explain this
- The play time is tracked as a `Tracker.Data.playtime` stat, which is saved as part of the tracked data
- The play time also displays on the Stats screen and Game Over screen
   - The Timer is paused on the Game Over screen, and unpaused when the player clicks Continue or Retry Battle.
- For ease of implementation, the Timer's time tracking isn't very granular; it simply counts seconds-elapsed using `os.time()`
   - Because it uses `os.time()`, it's possible this doesn't work for Linux or Mac OS. I removed my VMs, so I wasn't able to test.

image


ironmon_tracker/Program.lua
```diff
@@ -31,6 +31,97 @@ Program.GameData = {
...
+		local clicked = Input.isMouseInArea(xmouse, ymouse, self.box.x, self.box.y, self.box.width, self.box.height)
+		if clicked then
+			if self.isPaused then
+				self:unpause()
+			else
+				self:pause()
+			end
+			self.readyToDraw = true
+		end
+	end,
```
Fellshadow, line 104: This part doesn't check for if the play time is being displayed, meaning if someone happened to click down where the timer would be (can potentially happen by accident if they have another window over bizhawk for example) it will be paused in the background and if they decide to show it later in the run it won't have actually been updating

Fellshadow (changes requested):
Looks good to me, only issue I have is that you can pause the timer in the background while it's not being displayed (more details in the comment below)

Other comment is just a curiosity really, not something I have issue with

ironmon_tracker/Utils.lua
```diff
@@ -97,6 +97,17 @@ function Utils.formatControls(gbaButtons)
 	return controlCombination:sub(1, -3) or ""
 end
 
+function Utils.formatTime(numSeconds)
+	local sec = numSeconds % 60
+	local min = math.floor(numSeconds / 60) % 60
+	local hour = math.floor(numSeconds / 3600) % 100
```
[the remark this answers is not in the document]
UTDZac: Yeah, its just an arbitrary choice. I suppose its possible to go for several hundred hours. I mostly wanted some cut-off here in the event there was a weird glitch with retrieving the playtime numSeconds and the result was a stupidly large number.

I'll update this from `% 100` to `% 10000` to allow for hours up to 9999. That should be sufficient and still meet my goal.

ironmon_tracker/Program.lua
UTDZac: Fixed with https://github.com/besteon/Ironmon-Tracker/pull/391/commits/6c6c76e237f83359055965313292a694cd6cb2e5

Fellshadow (approved):
Looks good

PR #392, opened 2023-07-16
Crash Recovery Feature and Game Timer Updates

UTDZac:
This PR adds a new Crash Recovery feature to help recover game saves whenever an emulator crash occurs. This is done _not_ by saving information at the moment of the crash but rather by frequently creating a persistent backup save file. These save files are stored in a new folder within the Tracker folder such that they are easy to locate.

This is a simple feature in concept yet I imagine it will be immensely helpful for several people, especially for people like me who forget to save their progress often. 

### Crash Recovery Feature Notes
- Compatible with Bizhawk and mGBA emulators. The UI only exists for Bizhawk however.
- Backup save files for the current game are stored in the `backup_saves` folder.
   - An additional backup save is also kept for the previous game, saved right before the Quickload command is executed
- When a crash\* occurs, Bizhawk will detect that it did not shut down properly. When it does, it will show the Crash Recovery screen when the Tracker starts up to let the player know a crash occurred, then ask if they want to try and recover the save.
   - You can also navigate to this screen at any time from Settings -> ExtrasScreen.
- On the Crash Recovery screen, you can restore the game using the most recent backup save file. After you do so, the button swaps to an "Undo" button which will undo the load save file action.
- The recovery screen also provides some details about the state of the game you were playing when the crash occurred, comparing them to the game being played now. This is to help ensure you load the same game. This isn't very exact or foolproof, but it should help identify which rom was loaded.
- Added a `folderExists` function to FileManager, likely useful for other features
- Minor modifications to other screens to make room on the ExtrasScreen for this feature

\* By "crash" I believe only applies when the emulator itself runs into an error that causes it to force quit, or it freezes, or there is a power outage. A "crash" won't occur if the player simply closes the emulator, or closes the lua console, or "Ends Process" through Task Manager. Regardless of a crash occurring, the backup save files are always kept up to date, but the Tracker detecting the crash and prompting the player might not always occur.

image

### Game Timer Updates
In preparation for the upcoming release, I also wanted to add in some requested changes to the onscreen Game Timer feature. Doing so required modifying the shared space on the ExtrasScreen. As such, I bundled these changes with the Crash Recovery feature for convenience. Apologies if this makes the review process trickier.

- You can now edit the time of the game timer, done through a popup prompt
- You can now change where the game timer is shown on the screen.
   - Choice of six locations: the four corners, top-center, or bottom-center.


Fellshadow (approved):
Looks good to me

PR #398, opened 2023-07-18
Patch 8.0.1

UTDZac:
I also fixed the timing of when release notes were shown on screen. They now appear **after** a successful update to a new version, instead of before.

I also updated the move description for Shield Dust. I don't think this or the above change need to be mentioned in the release notes though.

Release notes to include with the 8.0.0 notes
- [v8.0.1] Fixed an issue when using both on-screen timer and repel usage
- [v8.0.1] Fixed an issue with move names being invisible on the Move History screen


Fellshadow (approved):
Looks good to me

PR #399, opened 2023-07-18
Patch 8.0.2

UTDZac:
Like clockwise, there's always a few things missed. Luckily this one was also an easy fix.

### Release Notes
- [v8.0.2] Fixed an issue where the Crash Recovery screen would appear after a successful Tracker update


Fellshadow (approved):
👀

PR #400, opened 2023-07-19
Patch 8.0.3

UTDZac:
Somehow an extra space snuck in to two recent screen files, which caused the files to file to load on Mac and Linux machines. This fixes that problem (someone helped me test it)

I also made that minor Resource change for the "minute" vs "minutes" language entry. I tested this briefly on the Time Machine screen and it worked fine. No need to mention in the release notes, as its a non-functional change.

### Release Notes
- [v8.0.3] Fixed a critical bug that prevented the Tracker from loading on Mac and Linux


jtigues (approved):
Looks good

PR #403, opened 2023-08-09
Improved Log Viewer with Routes!

### LINKED ISSUE #357: Log Viewer Improvements Task List
Just writing down some of the things that I've wanted to add to the log viewer and the parsing method to make it more compatible with how people randomize their Pokémon games.

- [x] Parser supports custom names (Pokémon, moves, etc.) instead of hard-coded values
- [x] Parser should read in and use the ability data, instead of relying on the data from the currently loaded game.
- [x] Route menu to check what wild Pokémon are on each route. Perhaps have an actual map shown here
- [ ] Fun stats added to the Misc. tab, stuff like average BST of the elite four.
- [x] A way to "search" the log for something. Something like "where are all the Shedinjas," "which Pokémon have Perish Song," and "what did TRAINER_NAME_HERE have"
- [x] A way to quickly view the log file for the previous seed after resetting (from Extras screen)

UTDZac:
This feature extends the usability of the Log Viewer and Log Search of the Tracker.

These additions should close #357 , by checking off the remaining items from the feature list

### What's New
- All routes (map locations) for RSE and FRLG have been added to the Log Viewer
   - _Note: Ruby, Sapphire, and Emerald were a nightmare to get accurate. Likely some incorrect manual data entries._
- Each route has info on all wild encounters and trainers, separated accordingly
- The Log Viewer tabs are now icons, which change occasionally on the game being played
- Trainers displayed in the Log Viewer are shown with clean portrait images and their team count displayed below them
- More contextual search options
   - For example, you can search for any route by its name, or trainer name, or Pokémon name.
   - Or search for routes with trainers that have a specific move or ability amongst their team of Pokémon
- Search results will highlight on their respective screens, making it easier to find what you've searched for
- Added an option to Log Viewer Misc. tab for displaying Custom Trainer Names / Classes
- **[Internal]** Cleaned up a ton of the Log Viewer code files, splitting off the massive singular file into more manageable tab screen files.

### Searching all Routes
image

### Viewing a Single Route Location
image


ironmon_tracker/data/TrainerData.lua
```diff
@@ -75,17 +124,329 @@ function TrainerData.initialize()
...
+		for i=492, 497, 1 do
+			trainerIds[i] = true
+		end
+		for i=511, 515, 1 do
+			trainerIds[i] = true
+		end
+		trainerIds[147] = true
+		trainerIds[200] = true
+		trainerIds[263] = true
+		trainerIds[530] = true
```
Fellshadow, line 427: Perhaps a cleaner / more compact way of defining these could be to define the ranges of dummy trainer ids and then looping through to define the `[id] = true` array items, e.g. something like:
```lua
local idRanges = {
    { 40, 43 },
    { 47, 50 }.
    ...
    { 851, 855 },
}
for _, range in pairs(idRanges) do
    for i=range[1], range[2], 1 do
        trainerIds[i] = true
    end
end
```
of course this table of ranges could be written out however you feel looks best, and any specific single IDs that don't lie within a range could be added after this loop *or* in the `idRanges` as `{ id, id }`

Fellshadow (commented):
Haven't had a chance to load it up and test it out on my end, but didn't see any huge issues in the code so take this as a "soft" approve I guess lol

Left a couple of small comments, just suggestions to think about

ironmon_tracker/data/TrainerData.lua
UTDZac: This is a good suggestion. I ended up doing just this in one of the later additions to a long list of items, as seen in the Ruby/Sapphire trainer mappings in the same file.

I'll get this included.

UTDZac: Resolved https://github.com/besteon/Ironmon-Tracker/pull/403/commits/da72c7752034b834bf6730af2b2b11957766894c

ironmon_tracker/FileManager.lua
```diff
@@ -125,6 +126,14 @@ FileManager.LuaCode = {
...
 	{ name = "LogOverlay", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogOverlay.lua", },
+	{ name = "LogTabPokemon", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabPokemon.lua", },
+	{ name = "LogTabPokemonDetails", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabPokemonDetails.lua", },
+	{ name = "LogTabTrainers", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabTrainers.lua", },
+	{ name = "LogTabTrainerDetails", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabTrainerDetails.lua", },
+	{ name = "LogTabRoutes", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabRoutes.lua", },
+	{ name = "LogTabRouteDetails", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabRouteDetails.lua", },
+	{ name = "LogTabTMs", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabTMs.lua", },
+	{ name = "LogTabMisc", filepath = FileManager.Folders.ScreensCode .. FileManager.slash .. "LogTabMisc.lua", },
```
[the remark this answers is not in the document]
UTDZac: I think I'm okay with just leaving all these screen related coded files in the `screens` folder as-is. Especially since they are already nicely categorized with "LogTab" as a prefix.

UTDZac:
I snuck in a "no results found" when searching for something results in nothing being displayed. Figured this little Easter egg was more helpful in describing the search output than just emptiness.

ironmon_tracker/data/TrainerData.lua
```diff
@@ -75,17 +124,113 @@ function TrainerData.initialize()
...
+			{1, 88}, 147, 200, 263, {492, 497}, {511, 515}, 530,
+		}
+	end
+
+	local trainerIds = {}
+	for _, item in pairs(trainerIdRanges) do
+		if type(item) == "number" then
+			trainerIds[item] = true
+		elseif type(item) == "table" then
+			for i = (item[1] or 9999), (item[2] or 9999), 1 do
```
Fellshadow, line 187: do we need the fallback defaults here, considering that the ranges are just defined right above? (so we know there *shouldn't* be any nil or missing items unless there was user error in whomever is editing the `trainerIdRanges`)

In this case, if we're factoring in possibilities for dev error on the array defined above, there could technically be an issue where if `trainerIdRanges` contains any value like `{ idX }` (i.e. is a table but with one value) then we wind up filling in values for `trainerIds` from `idX` to `9999` (including trainers that we wouldn't want to exclude) which may / may not be as obvious initially as having an error from nil / missing value

ironmon_tracker/data/TrainerData.lua
```diff
@@ -75,17 +124,113 @@ function TrainerData.initialize()
...
+	for class, trainers in pairs(classMap) do
+		for _, item in pairs(trainers) do
+			-- Could be a list of raw ids, or a range ids as pair (fromID, toID)
+			if type(item) == "number" then
+				trainerList[item] = {
+					class = class,
+					group = class.group
+				}
+			elseif type(item) == "table" then
+				for i = (item[1] or 9999), (item[2] or 9999), 1 do
```
Fellshadow, line 225: See other comment above about the fallback defaults

Fellshadow (commented):
Seems fine to me, just one comment about a particular usage of default fallback values that may be worth considering before merging in

ironmon_tracker/data/TrainerData.lua
UTDZac: Good catch, and yes that makes total sense. I'll remove the defaults and just add a slightly safer check to the entry condition.

Fixed in https://github.com/besteon/Ironmon-Tracker/pull/403/commits/2b90717aaba420a65feeedf24abcc4e88f826be0

Fellshadow (approved):
yeah that seems good to me now

PR #404, opened 2023-08-11
Open Book play mode feature

UTDZac:
Added Open Book play mode. Enable this log feature to see log info while actively playing the game. The option lives in the Log Viewer's Misc tab option list, since it's more-or-less part of the log.

_(This feature requires the extended Log Viewer routes feature, that PR must be merged in first)_

### Info Revealed by Open Book
- Wild and Trainer Pokémon's stat boxes are replaced with base stats (the bst spread, not actual in-game stats)
   - These values are highlighted to help further indicate they are revealed log info, and not true in-game stats
- Pokémon abilities
- Pokémon moves
- The entire Pokémon moveset is shown in the Move History screen
- The Route info page that normally tracks wilds you encounter now just shows all wild encounters in that area, with their levels & percentages

### Quick Access to the Log
Added a magnifying glass button in the space where the survival PC heals get shown on the main Tracker screen. Clicking it opens the log viewer. Also, clicking on the Pokémon icon and/or stats area of an enemy Pokémon will opens the log viewer for that Pokémon, offering quick-access to the log.

image

### Feature Complications
To get this feature to work, I told the Tracker to parse the log file when it starts up. This adds a bit of time to startup, since parsing the log can take a second or two. Don't really see a way around this, since you need access to all that info in order to display it while actively playing. If no log is auto-detected (from Quickload settings) to be parsed, then it prompts the user to open it. Again, no good way around this. As a result of this, I added a "warning" to the Open Book option to help inform the player that some extra stuff has to happen in the backend, and on their part, to make this feature work.

openbook-feature

### DIFF
--- a/ironmon_tracker/Languages/English.lua
+++ b/ironmon_tracker/Languages/English.lua
@@ -510,7 +510,7 @@ ScreenResources{
 		CheckboxShowPreEvolutions = "Show Pre Evolutions",
 		CheckboxCustomTrainerNames = "Custom Trainer Names",
 		CheckboxOpenBookMode = "Open Book Play Mode",
-		LabelExtraTimeWarning = "Adds 2 seconds to game load",
+		LabelExtraTimeWarning = "May add extra load time",
 		PromptShareSeedTitle = "Share Randomizer Seed",
 		PromptShareSeedDesc = "Copy/paste everything below to share. Load it through Randomizer --> Premade Seed.",
 	},
--- a/ironmon_tracker/Languages/English.lua
+++ b/ironmon_tracker/Languages/English.lua
@@ -509,6 +509,8 @@ ScreenResources{
 		ButtonShareSeed = "Share Seed",
 		CheckboxShowPreEvolutions = "Show Pre Evolutions",
 		CheckboxCustomTrainerNames = "Custom Trainer Names",
+		CheckboxOpenBookMode = "Open Book Play Mode",
+		LabelExtraTimeWarning = "Adds 2 seconds to game load",
 		PromptShareSeedTitle = "Share Randomizer Seed",
 		PromptShareSeedDesc = "Copy/paste everything below to share. Load it through Randomizer --> Premade Seed.",
 	},
--- a/ironmon_tracker/Tracker.lua
+++ b/ironmon_tracker/Tracker.lua
@@ -410,9 +410,7 @@ end
 
 -- If the viewed Pokemon has the move "Hidden Power" (id=237), return it's tracked type; otherwise default type value = NORMAL
 function Tracker.getHiddenPowerType()
-	-- Default to 'true' if not using open book setting
-	local useOpenBookInfo = not Options["Open Book Play Mode"] or Tracker.Data.isViewingOwn
-	local viewedPokemon = Battle.getViewedPokemon(useOpenBookInfo)
+	local viewedPokemon = Battle.getViewedPokemon(true)
 
 	if viewedPokemon ~= nil and Tracker.Data.hiddenPowers[viewedPokemon.personality] ~= nil then
 		return Tracker.Data.hiddenPowers[viewedPokemon.personality]
--- a/ironmon_tracker/Tracker.lua
+++ b/ironmon_tracker/Tracker.lua
@@ -410,7 +410,9 @@ end
 
 -- If the viewed Pokemon has the move "Hidden Power" (id=237), return it's tracked type; otherwise default type value = NORMAL
 function Tracker.getHiddenPowerType()
-	local viewedPokemon = Battle.getViewedPokemon(true)
+	-- Default to 'true' if not using open book setting
+	local useOpenBookInfo = not Options["Open Book Play Mode"] or Tracker.Data.isViewingOwn
+	local viewedPokemon = Battle.getViewedPokemon(useOpenBookInfo)
 
 	if viewedPokemon ~= nil and Tracker.Data.hiddenPowers[viewedPokemon.personality] ~= nil then
 		return Tracker.Data.hiddenPowers[viewedPokemon.personality]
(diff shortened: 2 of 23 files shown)


UTDZac:
This also makes a few tiny changes to the Log Viewer itself, such as showing the route info on the main tracker screen when viewing a single route on the log viewer screen. This technology didn't exist before, as I had to add in updates to make the route info tracker screen reveal all the game's current Pokémon. With this inclusion, I added back in the selection box around the log viewer route screens encounter list, so its more clear that you know which encounter you're viewing and that you can swap to view a different one.

image

ironmon_tracker/Languages/English.lua
```diff
@@ -509,6 +509,8 @@ ScreenResources{
 		ButtonShareSeed = "Share Seed",
 		CheckboxShowPreEvolutions = "Show Pre Evolutions",
 		CheckboxCustomTrainerNames = "Custom Trainer Names",
+		CheckboxOpenBookMode = "Open Book Play Mode",
+		LabelExtraTimeWarning = "Adds 2 seconds to game load",
```
Fellshadow, line 513: (This is the physicist in me coming out here) In what way was this 2 seconds measured, is it an estimate from one system or more an average observed across multiple?

Considering the huge variance of systems out there I feel it could be tough to quantify for certain, may be better to word it as something more vague like "Loading on new game will take slightly longer" (or whatever wording fits)

ironmon_tracker/Tracker.lua
```diff
@@ -410,7 +410,9 @@ end
 
 -- If the viewed Pokemon has the move "Hidden Power" (id=237), return it's tracked type; otherwise default type value = NORMAL
 function Tracker.getHiddenPowerType()
-	local viewedPokemon = Battle.getViewedPokemon(true)
+	-- Default to 'true' if not using open book setting
+	local useOpenBookInfo = not Options["Open Book Play Mode"] or Tracker.Data.isViewingOwn
+	local viewedPokemon = Battle.getViewedPokemon(useOpenBookInfo)
```
Fellshadow, line 415: I'm a bit confused by the variable naming here of `useOpenBookInfo`, it's set to true if the option for open book is disabled / we are *not* using open book?

Also a bit confused as to what this bit of code is doing overall here as opposed to what the removed line was doing, in regards to the desired behaviour both with / without open book setting enabled

ironmon_tracker/data/DataHelper.lua
```diff
@@ -226,7 +250,7 @@ function DataHelper.buildTrackerScreenDisplay(forceView)
 
 		-- Update: Specific Moves
 		if move.id == 237 then -- 237 = Hidden Power
-			if data.x.viewingOwn then
+			if data.x.viewingOwn or useOpenBookInfo then
 				move.type = Tracker.getHiddenPowerType()
```
Fellshadow, line 254: Not sure why we are using `useOpenBookInfo` for hidden power type?

This functionality is for showing the user's manually-tracked hidden power type, or if there is none it returns the same `MoveData.HiddenPowerTypeList[1]` that the `else` block here does too.

If we're planning to let people manually set a type for enemy pokemon that doesn't seem like something to limit to the open book setting? Though realistically, in gen 3, we know trainers mostly have no IVs so like 90+% of trainers will have hidden power fighting anyways. If it were the case that they had varying IVs and the intent here would be to show the resulting hidden power type, that'd be info that you don't get from a log file (since it derives from IVs)

Fellshadow (commented):
Mostly fine to me, just a couple questions

ironmon_tracker/Languages/English.lua
UTDZac: I picked "2 seconds" as just a way to use less words to convey the meaning that enabling this option will add some extra startup time. I had trouble finding words that'd fit, especially if this will be translated.

2 seconds is more than 1 second, which means the extra time is not nothing. In reality (for my machine), it takes 1 second or less. So the "2 seconds" is more of a, if it goes faster thats great, but do expect some delay.

I'm up for using a different descriptive warning message here.

ironmon_tracker/Tracker.lua
UTDZac: The labels here are rough, but ultimately its a change from only ever showing hidden power type for *your* pokemon, and allowing it to show the hidden power type of the enemy pokemon.

Which I only just now realized doesnt actually show the hidden power type as defined by the game, and just shows the type you've tracked. Whoops. I'll revert this change.

ironmon_tracker/data/DataHelper.lua
UTDZac: Fixed this in https://github.com/besteon/Ironmon-Tracker/pull/404/commits/9bfba4ad6e24adcd1f9adc5ae1f6d08287c8d433

ironmon_tracker/Languages/English.lua
Fellshadow: fair, though the thing to keep in mind is that different people have different thought processes and interpretations of things. I think it'd be better to not give specific amounts (unless they can be quantified as such) and stick to a more general warning that it may increase startup time slightly, since that's ultimately what we're wanting them to expect

ironmon_tracker/Tracker.lua
Fellshadow: Yeah the latter is why I was confused about the whole usage, though even if it did work like that i'd still feel unsure about showing hidden power types given that it's derived from IVs (and not info from the log), as mentioned in the other comment

ironmon_tracker/Languages/English.lua
UTDZac: How's this?

image

Fellshadow: hm, the "game load" part feels a little off (since it's technically the tracker loading not the game)
Does something like "May add extra load time" fit / work?

UTDZac: I like that more, committed that. https://github.com/besteon/Ironmon-Tracker/pull/404/commits/30230de2e7fdc1b7f1700667c3cb235c47e8aaa8

Fellshadow (approved):
yup, all good to me now

PR #407, opened 2023-08-25
Adding new address

jtigues:
Fixing bug where moves were tracked in the small window of wake-up text, even if some other status condition i.e. Confusion prevented them from using the move.


UTDZac (approved):
Looks good. Thanks for the quick fix.

PR #408, opened 2023-09-01
Compacting the PC Heals / More space for "Heals" translations

Fellshadow:
- Replaces the "PC Heals" header and toggle checkbox with a heart icon that can toggle the auto-counting of PC Heals
- Shortens "Heals in Bag" to just "Heals" for shorter translations to other languages

Note, I see there is an existing translation for Japanese for "Heals in Bag", since I don't know Japanese I didn't want to touch this translation as I don't know what the equivalent to just "Heals" would be for shortening it

Toggle off:
image

Toggle on:
image


UTDZac (approved):
Looks great. Thanks for finding more room here.

PR #409, opened 2023-09-03
Animated Iconset: Walking Pals

UTDZac:
This new iconset introduces new Tracker feature support for animation. Normally animated gifs can't be drawn on Bizhawk, but this gets around that by manually animating images with their image files and respective frame data.

### Walking Pals Iconset
- This is the 6th iconset to be added to the Tracker
- This iconset is animated and interactive! The icons will respond to in-game events such as walking, sleeping, or fainting
   - When walking around the map, the icon will face the same direction as the player
   - When put to sleep in battle, the icon will use a sleeping animation
   - For some Pokémon, when they faint in battle, the icon will use a fainting animation
   - When the player is afk, the icon will fall asleep out of boredom
- An option to disable "walk with me" is available on the Setup screen for the iconset, labeled as "Walk"

WalkingPalsPreview

### Additional Changes
- Added support for future animated iconsets by defining an `iconKey` with the iconset
- Added a generic `Program.createFrameCounter(N, func)` function to make it easier to setup a recurring function `func` to be executed once every `N` frames.
- For many text-areas in the Log Viewer, they are now drawn with a semi-transparent background to accommodate oversized animated sprites overlapping them.


ironmon_tracker/Program.lua
```diff
@@ -416,6 +441,48 @@ function Program.stepFrames()
...
+		step = function(self)
+			if self.paused then return end
+			-- Sync with client frame rate (turbo/unthrottle)
+			local delta = scaleWithSpeedup and (1.0 / Program.clientFpsMultiplier) or 1
+			self.framesElapsed = self.framesElapsed + delta
+			if self.framesElapsed >= self.maxFrames then
+				self.framesElapsed = 0.0
+				if type(callFunc) == "function" then
+					callFunc()
+				end
```
Fellshadow, line 476: with a variable name like `callFunc` I imagine the intent is *always* to pass in a function?

In which case, since this is an internal function only ever used with input arguments defined by ourselves, I feel like it would make more sense to check if `callFunc` is a function at the start of this function and if not then output some sort of error message for the dev to say "hey, this expects a function"?

My point here being that something within our control where an input argument can only be incorrect *if* a dev has incorrectly called this function **should** bring up an error and not obfuscate the incorrect usage behind a failsafe, the concern being that the latter naturally lets the dev be more lax in ways that can let bugs slip through

I'd also argue the same regarding `label` and `frames`, if the values for those are incorrectly passed in then it's the fault of the dev and they should be informed via an error message about it.

Fellshadow (commented):
Certainly a great addition, myself and many others are big fans of the mystery dungeon games so I know a lot of people would love this :)

Only thing that I think may be worth discussing / thinking about is the line being blurred between "fail-safes in case of bad memory reads / data outwith our control" and "fail-safes in case a dev assigns a bad value to a variable or passes in an incorrect argument" (not including "default values" for function input arguments here, those are fine and pretty standard).
In the latter case, where the dev has full control over the variables and function arguments, imo it should be as obvious as possible when a bad value is used so we can catch as many mistakes/bugs as we can in testing rather than obfuscating the errors with fail-safes.

Other than that, didn't spot any glaring issues in the code

ironmon_tracker/Options.lua
```diff
@@ -53,49 +54,58 @@ Options = {
...
-		author = "Ryastoise",
+	},
+	{
+		name = "Walking Pals",
+		author = "UTDZac",
+		folder = "spritesWalkingPals",
+		extension = ".png",
+		isAnimated = true,
+		iconKey = "WalkingPals",
 	},
```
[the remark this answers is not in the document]
UTDZac: Good note. I'll add in a `source` property for the ones that I can find. I'm going to comment-out this property (save space), since we don't have a use for it yet. I imagine it will be easy to add in some external link to open up the source later; at which time we can uncomment these.

ironmon_tracker/Program.lua
UTDZac: Updated the internal documentation for this function such that it is significantly more clear how it's expected to be used, and that the `callFunc` is an optional parameter.

Fellshadow (approved):
seems good to me 👍

PR #412, opened 2023-09-15
8.3.0 Release

UTDZac:
If/when you have time, can you double-check your Quickload stuff still work? I'm fairly confident the small checks I did to load in the proper attempts count should work fine, but also worried I missed something. Also try changing to another quickload setup you've used in the past, such as switching games or challenge modes (rnqs settings file). Hopefully the attempts counter stuff updates itself properly and the quickload screen UI properly reflects the changes. 

### Release Notes
- You can now look up all possible [Randomized Evos](https://user-images.githubusercontent.com/4258818/266699708-174a0b8a-281d-459a-bc9e-129c8c66f18c.png) by clicking on a Pokémon's icon; percentages shown are estimates (Shoutout to brdy and Harkenn)
- A simple [Coverage Calculator](https://user-images.githubusercontent.com/4258818/266700515-e001d5ec-f9bf-4c6a-99c6-b01609e2c31a.png) has been added under Extras > Tools
- Fixed a critical bug where changing Quickload settings would incorrectly overwrite your attempts count. In addition, the Quickload screen setup UI has been updated to help show what settings you are using
- Fixed a bug where manually loading saved Tracker notes would fail to actually load them into the Tracker
- Fixed a bug where the "Congratulations" screen would change to "GameOver" after leaving the Hall of Fame
- Fixed a bug where the drop-down list of Themes would grow in size
- For a full list of new features, changes and bug fixes, check out the **[Version Changelog](https://github.com/besteon/Ironmon-Tracker/wiki/Version-Changelog)**

#### Quickload Related changes
- 4fc34e731e2ea1925bf40f77fe2d4d2f8940e0bd Changing quickload properly updates attempts
- edc2d36d19cfd1894e1861441cf3eab7141ebea5 Fix Attempts file incorrect reference
- 701ce3bd22f0a1d4075cab43664f041bdc147d82 Update attempts # when swapping methods
- 6f5f450452a32413b147c552e996bb610a92e413 Quickload Screen UI updates

#### Other changes not included in a PR
- 87b419da8e275538a315c15e47b70f07d904bd71 Add in remaining base friendship values
- e39876426a16d9e91823ec4d01a331272d2d5094 Load Tracked Data now force imports
- 13e8535322c88570e4f311f75c37f9ed8661ad15 Add checks for Shiny and PokéRus
- 5d4e74c90cbd35410dec6033cca1c9a741cb6bd2 Fix CONGRATS screen from changing after win
[...]

PR #413, opened 2023-09-15
Fix for new quickload attempts

Fellshadow:
When testing the quickload changes included in #412 (will want to merge this before that PR) I noticed consistently that swapping to a new settings file for quickload *while* a tracker-generated rom is currently loaded results in the tracker erroneously carrying over the attempts file from the previous quickload settings file into the new one

Example reproduction steps:
1. Set up tracker quickload for one particular settings file e.g. `RSE Kaizo`
2. Quickload a couple of roms / edit the attempts count to some arbitrary number > 1
3. Change the quickload settings file to another settings file e.g. `RSE Survival` while `RSE Kaizo AutoRandomized.gba` is still loaded in the emulator
4. Notice how the attempts counter from `RSE Kaizo Attempts.txt` carries over into `RSE Survival Attempts.txt`

Went with adding an optional (default false) parameter to force the attempts filepath to be based off the currently set settings file & create a new attempts counter for it if it does not exist. Seems to work fine for me in the above steps but definitely test in case there's a case I didn't catch / it broke elsewhere

### DIFF
--- a/ironmon_tracker/Main.lua
+++ b/ironmon_tracker/Main.lua
@@ -783,7 +783,12 @@ function Main.SaveCurrentRom(filename)
 	return nil
 end
 
-function Main.GetAttemptsFile()
+--- Attempts to find an attempts file based off of the Quickload settings file, then off of the currently loaded ROM name
+--- @param forceUseSettingsFile boolean Force return of filepath of an attempts file to be created from the Quickload settings file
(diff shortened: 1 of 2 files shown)


UTDZac (approved):
Reviewed the code changes, everything looks good. These changes properly take care of the scenario where no attempts count exists for a given settings file setup, and thus resets it back to default.

PR #415, opened 2023-09-24
Tracker Data Restructure and Optimization

UTDZac:
This PR primarily focuses on changing the structure of the internal `Tracker.Data` and external `.TDAT` notes file. The goal being to track less data and optimize for speedy gameplay (large end-game TDAT files). In addition, I've included other useful data from the game for future projects.

Hopefully I did this all correctly such that old TDAT files are compatible with a new Tracker version, assuming it gets upgraded in the middle of a game / between play sessions. I did a good bit of testing with existing log files, but would appreciate additional testing if you can afford it.

As such, I purposefully changed the Tracker version to **v8.3.5**, not with the intent of doing a release, but as a safeguard for anyone that wants to use/test with it and try to use old TDAT files.

### New Structure
In an effort to help better structure what the commonly used data objects for a "Pokémon" look like, I've changed up the simple table structure to what is effectively a "class object" for Lua. This is accomplished through [default object tables and metatables](https://www.lua.org/pil/16.1.html).
- `Program.DefaultPokemon` is a "Pokémon" object that can be created
- Similarly, `Tracker.DefaultData` is what all new tracked data notes are created from

### Tracker Data Removal
A large part of this process was to identify what tracked data could simply just be removed, or moved elsewhere since they don't really need to be "tracked" between play sessions. This mostly applies to data such as a Pokémon object on a player's team. The data file could also be shrunk by unpacking some tables (less pickling to do), shortening variable names that appeared frequently, and leaving out 0 or default values that could otherwise be assumed (empty moves).

**For the data that are no longer tracked but simply moved**, that includes:
- Tracker.Data.ownPokemon -> Program.GameData.PlayerTeam
- Tracker.Data.otherPokemon -> Program.GameData.EnemyTeam
- Tracker.Data.isViewingOwn -> Battle.isViewingOwn

**For data that are still tracked but have been changed**, refer to `Tracker.checkForLegacyTrackedData()` to see how it now import an old TDAT file for compatibility

**For all other data that are removed**, no other changes are needed. The current Tracker simply won't import or use that data. Old TDAT files will have it temporarily until they're updated / overwritten.

### Additions
I started this PR with the intent of tracking what trainers the player defeats, for use in a new Tracker project in the near future. However, it turns out this information is just part of game memory itself and can be read in directly. I've included support for this through:

#### Info on Defeated Trainers
- `Program.hasDefeatedTrainer()` for checking a single trainer
- `Program.getDefeatedTrainersByLocation()` for checking a route or map location
- `Program.getDefeatedTrainersByCombinedArea()` for checking an entire "dungeon" or area with multiple floors
#### Info on TM Moves
- `Program.getMoveIdFromTMHMNumber()` looks up the moveId associated a given TM number, since this data wasn't previously available outside of parsing a log file


jtigues (approved):
Lot of stuff in here, looks solid to me.

UTDZac:
Fixed one last outstanding issue with defining default Tracker.Data attributes using `nil` as the default value (you can't do this btw, given the structure we're using). Thus, those attributes were never read in from a saved .TDAT file.

PR #418, opened 2023-10-02
Battle started flag

jtigues:
Added flags to give the tracker hard hooks for battle start, battle data being readable, and battle end.

- The conditions for being "In Battle" are still the same; as soon as the tracker finds an enemy mon in the enemy party data.
- Addresses added for the main battle function pointer, as well as several battle scripts relevant to key points in a battle.
- All delays and timers related to the start and end of battle have been replaced with checks against these scripts. A battle is considered "Started" (its data can be relied on) as soon as the pokemon begins sliding across the screen for Wild encounters, and immediately after the enemy's pokemon begins to appear from their pokeball for Trainer Battles
- Battle end is now the exact moment that the player returns from the battle to the overworld. As a bonus, this allows users that are still in the battle screen to view the enemy's last pokemon and mark stats etc, whereas before as sson as the enemy sprite came out the battle was already considered over.


UTDZac (approved):
Looks great. Thanks for solving this long overdue issue. I'm excited to try it out

PR #419, opened 2023-10-02
Fixing DataStart wild vs trainer distinction

jtigues:
Fixing issue where Wild and Trainer battles were looking at all the same script addresses, causing mons to be revealed earlier in Trainer battles.


UTDZac (approved):
Thanks for fixing it

PR #423, opened 2023-10-20
Battle started flag

jtigues:
- Emerald works a little differently due to how many "artificial" double battles can take place, and actually spends the bulk of its time in the BattleIntroRecordMonsToDex script. Previous address caused unreliable battle start times for some trainers since the script addresses from other games were only present for a few frames in Emerald.


Fellshadow (approved):
Seems good to me

PR #425, opened 2023-10-20
Add base EXP yield

Skargonas:
Feature Request #424

PR #426, opened 2023-10-20
Revert "Add base EXP yield"

UTDZac:
Reverts besteon/Ironmon-Tracker#425

Was merged into the wrong branch


jtigues (approved):
Fair enough

Issue #424, opened 2023-10-20
Add EXP yield


Skargonas:
### Tell us about your request

For Standard/Ultimate Ironmon players, it's useful to check the exp yield of Pokémon to know which to kill on each route.

### Additional Information

Already working on a patch that looks like this
![image](https://github.com/besteon/Ironmon-Tracker/assets/5890186/a7b01a95-43f3-473f-8d05-fdf730499d5e)

PR #427, opened 2023-10-21
Add base EXP yield

Skargonas:
FR #424


UTDZac (approved):
Thanks for doing this!

PR #429, opened 2023-10-24
Added an extra option to fast forward battle intro

mdmurphy2:
- Added an option under Extra Options that will automatically fast forward the intro sliding in / throwing pokemon scene
- This will Set speed to 6400 (max) and then reset back to what your client speed was set to
- Works on both wild interactions and trainer battles

### DIFF
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -804,7 +804,6 @@ function Battle.resetBattle()
 end
 
 --- Swap back to the main tracker screen if relevant. Handles a common case of looking up something on a different screen then starting/ending a battle.
---- @return boolean success True if successfully autoswaps; false otherwise
 function Battle.trySwapScreenBackToMain()
 	local screenAllowsAutoSwap = {
 		[InfoScreen] = true,
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -24,6 +24,9 @@ Battle = {
 	lastEnemyMoveId = 0,
 	enemyHasAttacked = false,
 	firstActionTaken = false,
+	
+	emulationSpeed = 100, -- (Percent) Used to store the last known client speed when fast forwarding battle intro
+	isFastForwarding = false,
 
 	-- "Low accuracy" values
 	Synchronize = {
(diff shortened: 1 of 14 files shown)


mdmurphy2:
image


Testimony xD

UTDZac:
It looks like this has some conflicts. I'd recommend trying to reset your branch to our current [dev](https://github.com/besteon/Ironmon-Tracker/tree/dev) branch, I'm not sure why this PR is including old stuff that got changed up.

This feature is one I want to explore more before committing to adding it. This is mostly a case of wanting to understand how it can be a helpful core addition to the Tracker (as opposed to a custom extension). I struggle to see a need or want for something like this. Who is this for? Typically people play the game without speedup at all, or with speedup permanently on, especially for early game.

If this is something we want to add, we'd need to resolve the below. Feel free to hold off on resolving these until we decide if this is something we want to add or not. I merely wrote them all down as I was testing and reviewing this feature.

A few notes about this feature:
- It doesn't play well if you play with sound on, which is common for those who don't use speed-up normally.
   - image
- It's a bit weird to me that the temp speedup hands at the "Wild PIKACHU appeared!" battle start dialog. I would expect the speedup to go away after the pokemon finishes sliding into view. I believe at one point we had the total number of frames it took for this to occur, so it might be as simple as just keeping speedup on for that many emulation frames (150 for wild pokemon, 80 for trainers... or those two might be swapped, its been a while)
- This will crash anyone who uses mGBA, since that emulator doesn't have Bizhawk's client functions. Ideally this feature can work regardless of what emulator you are using, otherwise, it needs proper checks to hide the option and not execute its code if on mGBA.
- In Battle.lua, the `emulationSpeed = 100` and `isFastForwarding = false` need to be set in the `initialize()` function of that file. The emulators have this weird thing where it remembers any current value defined in that global on reset/quickload. So if someone quickloads while the above variables have changed, their new game uses the changed values and not your expected defaults.
- Can you rename the option to "Fast forward battle intro" with lowercase letters to match other options
- Can you move the English language file option entry up a few lines so its grouped with other similar option labels
- When adding new text that appears on screen, you'll also have to add entries to the other language files available. Check the same area of other language files to see the template for how that looks.

UTDZac:
Closing, as this is more suited as a Tracker Extension than a core option.

PR #431, opened 2023-11-02
[draft] Enhanced per encounter details tracking

jwunderl:
Sounds like this will be better suited as an extension in case of perf issues on some devices, but putting up as a point of reference.

tracker-showing-prev-encounters


jwunderl:
Got it all moved over to https://github.com/jwunderl/EncounterDetails-IronmonExtension now so closing, thanks for the pointers on discord~

PR #432, opened 2023-11-02
8.3.8 Release

UTDZac:
Skipping from v8.3.6 to 8.3.8 due to beta testers being on 8.3.7 and I want them notified of this update (I'll be updating `beta-test` branch after this PR merger.

### Release Notes
- Fixed an issue where Game Over screen would appear at random times while playing Emerald, Ruby, or Sapphire
- Early pivot route info ("Walking Seen 2/5")  now shows outside of battle too
- Base EXP yield info now shown on the info page when clicking on the Pokémon icon. If you're currently in battle with that Pokémon, it will be the true EXP yield amount instead, based on the level of the Pokémon you're battling and shown in green
- Renamed "Quickload" to "New Runs", to help clarify the feature always loads a new game ROM and not an old ROM from a previous playthrough
- Fixed an issue where Pokémon names became fully capitalized if "Show nicknames" was enabled but no nickname was set
- Fixed an issue where the Update Screen would incorrectly notify of a new update when none was available
- Fixed move summaries for several multi-hit moves that were incorrectly stating the move made contact
- Coverage Calculator now automatically includes Hidden Power's type if it's set by the player
- Further internal data cleanup (Pokémon, Trainers, Battle stuff, etc) and optimization

Newest things to review if you want:
- 95d1675d40cbd61c658ae912121088d7375610a2 Early pivot route info, just walk around in early game areas (make sure disable mainscreen carousel setting is turned off to test this)
- c6815ed34822f58c0cafa8f5f1c15f06b567eadd Double-check the "New Runs" rename stuff works in other languages


ironmon_tracker/data/PokemonData.lua
```diff
@@ -447,3774 +396,2966 @@ PokemonData.TypeIndexMap = {
...
 Format for an entry:
+	pokemonID: integer -> The gen 3 pokedex id number for this Pokémon
 	name: string -> Name of the Pokémon as it appears in game
 	types: {string, string} -> Each Pokémon can have one or two types, using the PokemonData.Types enum to alias the strings
 	evolution: string -> Displays the level, item, or other requirement a Pokémon needs to evolve
 	bst: string -> A sum of the base stats of the Pokémon
-	yield: integer -> Base experience yield of the Pokémon
+	expYield: integer -> Base experience yield of the Pokémon
 	movelvls: {{integer list}, {integer list}} -> A pair of tables (1:RSE/2:FRLG) declaring the levels at which a Pokémon learns new moves or an empty list means it learns nothing
 	weight: pokemon's weight in kg (mainly used for Low Kick calculations)
```
Fellshadow, line 422: This "Format for an entry" comment is incorrect:
- `pokemonID`, `types`, and `yield`/`expYield` do not exist in the entries below
- `bst` is no longer a string in the entries below
- `weight` doesn't specify its datatype

ironmon_tracker/screens/TrackerScreen.lua
```diff
@@ -438,12 +438,35 @@ end
 
 -- Define each Carousel Item, must will have blank data that will be populated later with contextual data
 function TrackerScreen.buildCarousel()
+	-- Helper functions
+	local function isCarouselDisabled()
+		return Options["Disable mainscreen carousel"] and Battle.isViewingOwn
+	end
+	local function overrideEarlyRouteInfo()
```
Fellshadow, line 445: the naming of this isn't intuitive, at a first glance i took this to mean "returns true if we want to override showing the early route info and show something else" but instead it seems to be "returns true if we want to override what is currently there with the early route info"?
Perhaps something like "showEarlyRouteInfo" or "showEarlyRouteEncounters" could be a more intuitive name in that case

Fellshadow (changes requested):
Left a bunch of comments, but it's mainly additional bits of cleanup / clarifications / corrections

UTDZac:
Resolved all of your PR change requests, cleaned up comments, and fixed the merge conflicts with `main`

Fellshadow (commented):
Just one last little technicality, but rest seems good now

Fellshadow (approved):

PR #434, opened 2023-11-03
Patch 8.3.9

UTDZac:
### Release Notes
- [8.3.9] Fixed an issue with the Coverage Calculator incorrectly counting single-typed Pokémon as dual-type.
- [8.3.9] Fixed an issue where the step counter could get stuck at 0 steps at the beginning of a new game.

A note about the variables being reset: because they are defined outside of a function, Bizhawk does *not* properly reset them when a new game is started while the script is still running. They must be set in the init function to properly reset to default values.


jtigues (approved):
Looks good to me

PR #435, opened 2023-11-20
Add Network framework and Streamerbot support

UTDZac:
This PR adds a complete Network framework with connection support to [Streamerbot](https://streamer.bot/). For now, this is done simply by reading/writing every few seconds to inbound and outbound files (Json). Other connection types may become available in the future as they become officially (and properly) integrated into Bizhawk itself.

The initial goal of this feature is to allow Twitch chat to communicate with the Tracker via chat commands, or by redeeming a channel reward to change something about the tracker, like picking a starter ball or changing the Tracker theme.

### Setup
To user first needs to setup the Network through the **Steam Connect** screen found in the Streaming settings menu. This setup process will be outlined in a wiki page which I'll built out later. Once setup, the tracker will automatically connect when new games are loaded, and disconnect when the tracker resets/closes down.

- Tracker Wiki: [Stream Connect Guide](https://github.com/besteon/Ironmon-Tracker/wiki/Stream-Connect-Guide)

image

### Network Framework Details
- The **Network** class handles setting up connections, sending/receiving from the connection, and closing down the connections when necessary.
- The **RequestHandler** class manages incoming "Requests", either sent from Streamerbot or a Tracker extension, processes those requests, and creates "Responses" to send back. A Request needs a matching "Event" to determine how to process it and fulfill (complete).
- The **EventHandler** class contains all known "Events" that can be triggered. The type of events are chat commands and channel rewards, as well as internal game triggers which are currently only available through custom Tracker extensions. There are also core events with are managed here, such as Start, Update, and Stop events for establishing and syncing Network communication.
- To properly communicate with Streamerbot, C# code is required on be run on Streamerbot itself. This includes event entry points, such as chat commands and channel redeems, as well as all the other internal processes needed for sending requests and receiving responses.
- For external developers, I've added support to trigger on in-game conditions and send notifications to Streamerbot. A developer can build a Tracker extension which adds a new Game Event, then use Tracker code to determine when to trigger the event, which will send a request/response to Streamerbot. Then the developer can define their own Streamerbot client triggers to receive these notifications and respond accordingly, perhaps with a fun sound effect or game overlay update. Developing this will be explained in more detail on the Stream Connect wiki page.

### Connection Modes
- Text (Json) - Fully supported
- WebSockets - Not properly supported by Bizhawk
- HTTP - Not properly supported by Bizhawk

### New Core Files
- /network/Network.lua
- /network/RequestHandler.lua
- /network/EventHandler.lua
- /screens/StreamConnectOverlay.lua
- /Json.lua _(from a 3rd-party source)_
- /Debug/Tracker-StreamerBot.cs _(for streamerbot only)_

### Other Notes
- I accidentally added the `Utils.isNilOrEmpty()` helper function in this PR instead of the `dev` branch, which is why so many files are affected by this PR. Whoops!
- Ignore the Tracker version update, this PR will release with a new major version change anyway.
- Currently there are no Language resources for most of the new texts that were added. I'll add this in after any other code changes need to be made.
- I still want a "Delay Once" button for the pick ball redeems, that appears on the random ball picker. This is helpful for cases where a favorite or something pops up, and the user's ball redeem needs to be temporarily skipped.
- I still want to add a `!trainers` chat command for looking up what trainers the player still has left to fight. Having figured out the logic for this one yet. Will likely delay for a future patch release.
- There are a handful of other features and support for customizations, so let me know if something feels a bit off.

### How To Test
1. Install Streamerbot and log into the application
2. On Stream Connect, click "Get Code" and follow instructions for importing it
3. Then set the connection data folder through the tracker to the data folder found in Streamerbot.
4. Click Connect!

I'm interested in a better way to explain how to connect purely through the Tracker UI if at all possible. As of now, I plan to have instructions on the wiki. I think only step 3 is the hard part to explain that I'm stuck on.


UTDZac:
I've completed the [Stream Connect Guide](https://github.com/besteon/Ironmon-Tracker/wiki/Stream-Connect-Guide) wiki page if you want to check it out. The developers section is has a few links that won't work yet until this feature is officially released.

Fellshadow (approved):

UTDZac:
Finished adding in the resource files for this feature, as well as fixed a tiny bug with HMs displaying as TMs in chat output.

PR #438, opened 2023-11-26
Revert FileManager Utils changes

Fellshadow:
Tracker broke because of attempting to use Utils funcs before Utils is actually loaded, reverting use of funcs just in case


jtigues (approved):

jtigues:
Approved and merged, thanks.

PR #439, opened 2023-12-03
fix emerald route data pt1

CyanSMP64:
WIP of fixing emerald route data by adding missing level data and fixing incorrect stuff
so far i've stopped after cave of origin, so most of the maps are done but still some left to do


UTDZac:
Hi Cyan, would you be able to comment more on what this PR is about? For what purpose is it needed, what issue (if any) does it solve?

I noticed that it does add some useful wild encounter level data to `RouteData.lua`, which is great.

I'd be happy to work with you to understand more about the purpose of this PR and any questions with resolving conflicts. 

Unrelated to this, we're fixing some of the Emerald encounter data in an upcoming Tracker release later today.

UTDZac (approved):
Thanks for doing this. Reviewed both files and they look good to me to add in.

ironmon_tracker/data/RouteData.lua
```diff
@@ -4928,9 +4959,9 @@ function RouteData.setupRouteInfoAsRSE()
 			icon = RouteData.Icons.CaveEntrance,
 			area = RouteData.CombinedAreas.CaveOrigin,
 			[RouteData.EncounterArea.LAND] = {
```
CyanSMP64, line 4961: cave of origin b1f has no encounter data in emerald

PR #440, opened 2023-12-05
Retry Battle bugfix

jtigues:
Fixing a bug where when resetting a battle, enemy pokemon were not being read so moves/abilities were not tracked post-mortem


UTDZac (approved):
Thanks for finding this. Simple enough fix.

PR #442, opened 2023-12-20
Add Option to display Pokémon's gender

UTDZac:
Adds an Option to display a gender symbol next to a Pokémon's name. Works on both Bizhawk and mGBA emulators.

image

If the Pokémon's name or nickname is too long, the gender symbol will be displayed in the corner of the Pokémon icon.

image

The Option can be found under the Extras menu (really the only place that had room)


jtigues (approved):
Looks good

Fellshadow (approved):
looks good to me too

PR #453, opened 2024-04-15
Add files via upload

Zach0Dac:


Zach0Dac:
The color picker needs to be extended so translation fit more easily
Capture d’écran 2024-04-24 225226

Zach0Dac:
I can surely shorten those translations so they fit better, but I think they could all fit with some UI rework. Let me know if you would prefer or need me to find a shorter translation
image

PR #462, opened 2024-04-20
Better External UI framework

UTDZac:
This PR adds a long overdue wrapper for many of the [native Bizhawk Lua functions](https://tasvideos.org/BizHawk/LuaFunctions) for dealing with `forms`. Functionally speaking, nothing should change for the user outside of some form texts now actually being visible for non-English languages.

**The main purpose behind this was to make it easier to create and manage form popup windows**, including the individual Control elements associated with that form. The most notably change here is being able to easily apply a universal change to the creation of a given form Control, such as adding the property `"AutoSize"=true` on every Label used in all forms everywhere; great for translated text.

I did my best to document this new enhancement, update any old deprecated functions to use the new stuff, and test as many form popups as I could find on Bizhawk 2.8 and Bizhawk 2.9. Feel free to double-check me on any of these.

I'm also open to renaming the new `ExternalUI.lua` class to something shorter, if possible, since usage tends to get lengthy in some places. Note: This class has potential to be used for non-Bizhawk form things as well, as soon as we have a need for it.

### DIFF
--- a/ironmon_tracker/Theme.lua
+++ b/ironmon_tracker/Theme.lua
@@ -518,25 +518,27 @@ function Theme.openColorPickerWindow(colorkey)
 end
 
 function Theme.openImportWindow()
-	local form = Utils.createBizhawkForm(Resources.ThemeScreen.ButtonImport, 515, 125)
+	local form = ExternalUI.BizForms.createForm(Resources.ThemeScreen.ButtonImport, 515, 125)
+	if not form then return end
 
-	forms.label(form, Resources.ThemeScreen.PromptEnterThemeCode .. ":", 9, 10, 300, 20)
-	local importTextBox = forms.textbox(form, "", 480, 20, nil, 10, 30)
-	forms.button(form, Resources.AllScreens.Import, function()
+	form:createLabel(Resources.ThemeScreen.PromptEnterThemeCode .. ":", 9, 10)
+	local importTextBox = form:createTextBox("", 10, 30, 480, 20)
+	form:createButton(Resources.AllScreens.Import, 212, 55, function()
 		local formInput = forms.gettext(importTextBox)
-		if formInput ~= nil then
-			-- Check if the import was successful
-			local success = Theme.importThemeFromText(formInput, true)
-			if success then
-				Theme.refreshThemePreview()
-			else
-				print("Error importing Theme Config string:")
-				print(">> " .. formInput)
-				Main.DisplayError("The theme config string you entered is invalid.\n\nPlease enter a valid theme config string.")
-			end
+		if Utils.isNilOrEmpty(formInput) then
+			return
 		end
-		Utils.closeBizhawkForm(form)
-	end, 212, 55)
+		-- Check if the import was successful
+		local success = Theme.importThemeFromText(formInput, true)
+		if success then
+			Theme.refreshThemePreview()
+			Utils.closeBizhawkForm(form)
+		else
+			print("Error importing Theme Config string:")
+			print(">> " .. formInput)
+			Main.DisplayError("The theme config string you entered is invalid.\n\nPlease enter a valid theme config string.")
+		end
+	end)
 end
 
 function Theme.openExportWindow()
--- a/ironmon_tracker/Theme.lua
+++ b/ironmon_tracker/Theme.lua
@@ -519,12 +519,11 @@ end
 
 function Theme.openImportWindow()
 	local form = ExternalUI.BizForms.createForm(Resources.ThemeScreen.ButtonImport, 515, 125)
-	if not form then return end
 
 	form:createLabel(Resources.ThemeScreen.PromptEnterThemeCode .. ":", 9, 10)
 	local importTextBox = form:createTextBox("", 10, 30, 480, 20)
 	form:createButton(Resources.AllScreens.Import, 212, 55, function()
-		local formInput = forms.gettext(importTextBox)
+		local formInput = ExternalUI.BizForms.getText(importTextBox)
 		if Utils.isNilOrEmpty(formInput) then
 			return
 		end
@@ -532,7 +531,7 @@ function Theme.openImportWindow()
 		local success = Theme.importThemeFromText(formInput, true)
 		if success then
 			Theme.refreshThemePreview()
-			Utils.closeBizhawkForm(form)
+			form:destroy()
 		else
 			print("Error importing Theme Config string:")
 			print(">> " .. formInput)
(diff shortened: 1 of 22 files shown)


UTDZac:
Oh, the other reason I have been wanting to make this wrapper was because of the stupid inconsistency when using Bizhawk's Lua form function for creating a textbox. It is the *only* function that puts the `width` & `height` params *before* the `x` & `y` params. 😠

Fellshadow:
Taking a look but want to comment on this part first:
> I'm also open to renaming the new ExternalUI.lua class to something shorter, if possible, since usage tends to get lengthy in some places.

I think `ExternalUI` is a fine name as it's clear what it is, however instead of seeking a shorter name why not introduce a variable within the file to refer to `ExternalUI.BizForms` or whatever is being used in that file and use that instead to shorten the lengthy parts?

e.g. something like
```lua
local ui = ExternalUI.BizForms
```
and then anywhere you want to use `ExternalUI.BizForms` you do something like:
```lua
local form = ui.createForm(title, self.width, self.height, self.xPos, self.yPos, tryClose)
```
(obviously what variable names you use are up to you, but that seems the simpler way to shorten it and keep it clear imo)

ironmon_tracker/Theme.lua
```diff
@@ -618,18 +609,18 @@ function Theme.openSaveCurrentThemeWindow()
...
-			forms.settext(Theme.Manager.SaveNewConfirm, Resources.AllScreens.Save)
+			ExternalUI.BizForms.setText(Theme.Manager.SaveNewWarning, "Name cannot have 6 consectuive hexcode characters (0-9A-F)")
+			ExternalUI.BizForms.setText(Theme.Manager.SaveNewConfirm, Resources.AllScreens.Save)
 			return
-		elseif existingPreset ~= nil and forms.gettext(Theme.Manager.SaveNewConfirm) ~= Resources.AllScreens.Yes then
+		elseif existingPreset ~= nil and ExternalUI.BizForms.getText(Theme.Manager.SaveNewConfirm) ~= Resources.AllScreens.Yes then
 			-- If the Theme name is already in use, warn the user first
-			forms.settext(Theme.Manager.SaveNewWarning, "A Theme with that name already exists. Overwrite?")
-			forms.settext(Theme.Manager.SaveNewConfirm, Resources.AllScreens.Yes)
+			ExternalUI.BizForms.setText(Theme.Manager.SaveNewWarning, "A Theme with that name already exists. Overwrite?")
```
Fellshadow, line 622: should these warning texts be moved to resources language files to be translatable?

ironmon_tracker/Theme.lua
```diff
@@ -618,18 +609,18 @@ function Theme.openSaveCurrentThemeWindow()
...
-			forms.settext(Theme.Manager.SaveNewWarning, "Cannot use a reserved Theme name")
-			forms.settext(Theme.Manager.SaveNewConfirm, Resources.AllScreens.Save)
+			ExternalUI.BizForms.setText(Theme.Manager.SaveNewWarning, "Cannot use a reserved Theme name")
+			ExternalUI.BizForms.setText(Theme.Manager.SaveNewConfirm, Resources.AllScreens.Save)
 			return
 		elseif themeName:find("%x%x%x%x%x%x") then
 			-- Don't allow six consectuive hexcode characters, as this screws with parsing it later
-			forms.settext(Theme.Manager.SaveNewWarning, "Name cannot have 6 consectuive hexcode characters (0-9A-F)")
-			forms.settext(Theme.Manager.SaveNewConfirm, Resources.AllScreens.Save)
+			ExternalUI.BizForms.setText(Theme.Manager.SaveNewWarning, "Name cannot have 6 consectuive hexcode characters (0-9A-F)")
```
Fellshadow, line 617: typo here: `consectuive` -> `consecutive` (typo was already there but i guess wasn't noticed until now lmao)

ironmon_tracker/network/Network.lua
```diff
@@ -326,24 +326,24 @@ function Network.getStreamerbotCode()
...
 
 function Network.openUpdateRequiredPrompt()
-	local form = Utils.createBizhawkForm("Streamerbot Update Required", 350, 150, 100, 50)
+	local form = ExternalUI.BizForms.createForm("Streamerbot Update Required", 350, 150)
 	local x, y, lineHeight = 20, 20, 20
-	forms.label(form, string.format("Streamerbot Tracker Integration code requires an update."), x, y, 330, 20)
+	form:createLabel(string.format("Streamerbot Tracker Integration code requires an update."), x, y)
 	y = y + lineHeight
-	forms.label(form, string.format("You must re-import the code to continue using Stream Connect."), x, y, 330, 20)
+	form:createLabel(string.format("You must re-import the code to continue using Stream Connect."), x, y)
```
Fellshadow, line 333: Similar to other comment, should these labels be moved to the language files?

ironmon_tracker/screens/InfoScreen.lua
```diff
@@ -423,40 +413,30 @@ function InfoScreen.openPokemonInfoWindow()
...
-	local routeDropdown = forms.dropdown(form, {["Init"]="Loading Route Data"}, 50, 30, 145, 30)
-	forms.setdropdownitems(routeDropdown, RouteData.AvailableRoutes, false) -- true = alphabetize the list
-	forms.setproperty(routeDropdown, "AutoCompleteSource", "ListItems")
-	forms.setproperty(routeDropdown, "AutoCompleteMode", "Append")
-	forms.settext(routeDropdown, routeName)
-
-	forms.button(form, Resources.AllScreens.Lookup, function()
-		local dropdownSelection = forms.gettext(routeDropdown)
+	form:createLabel(Resources.InfoScreen.PromptLookupRoute .. ":", 49, 10)
+	local routeDropdown = form:createDropdown(RouteData.AvailableRoutes, 50, 30, 145, 30, routeName)
```
Fellshadow, line 437: the "false" to not alphabetize the list isn't provided in the new `createDropdown`, is it intentional to alphabetically order the route names here now?

Fellshadow (commented):
Looks mostly fine to me, just a few small comments

ironmon_tracker/Theme.lua
UTDZac: Technically yes, but I did not put all texts in the Resource file for translation help, mostly because it was just so many extra lines of texts to add. I wanted to just add the core texts to Resource files.

If/when this gets fixed, then it can be done in `dev` branch

ironmon_tracker/network/Network.lua
UTDZac: Not in this PR, see other comment.

ironmon_tracker/Theme.lua
UTDZac: I'll fix this in dev branch after merger.

ironmon_tracker/screens/InfoScreen.lua
UTDZac: Good catch. I'll fix this now

Fellshadow (approved):

PR #463, opened 2024-04-24
8.5.3 Patch Release

UTDZac:
Lots of little things have been fixed up and changed from user feedback. Thus, I want to get this seemingly small patch out and available now.

### Release Notes
- Added support for playing with champred's patch that changes all moves to use the Physical/Special split
   - View the online release notes for a patch download link: [Download v1](https://github.com/champred/pokeemerald/releases/tag/split-v1)
- OHKO move accuracy is now automatically calculated and shown
- Unlearnable Gym TMs in the Log Viewer are now more subtle. You can disable these from the Log Viewer options tab.
- [Ruby/Sapphire/Emerald] An icon showing that your Pokémon has PokéRus now displays on the Tracker screen
- Added a 1 second cooldown when using the New Run feature, to prevent it from occurring twice in a row
- Fixed an issue where the Tracker could not detect a battle had ended when playing on a French FireRed ROM
- Fixed some typos, as well as cleaned up several buttons and texts for popup windows


Fellshadow (approved):
Looks good to me

champred:
One small note for the physical/special split is that it is based off the most recent generation, so if any move categories changed it may not match up with what people are used to from the gen 4/5 tracker. May be good to mention, but it's probably not a big deal.

Fellshadow:
> One small note for the physical/special split is that it is based off the most recent generation, so if any move categories changed it may not match up with what people are used to from the gen 4/5 tracker. May be good to mention, but it's probably not a big deal.

from what i can see seems like there were only a few moves that changed category and they're later gen changes - https://bulbapedia.bulbagarden.net/wiki/List_of_modified_moves#Damage_category_changes

champred:
> spikes seems to be the only applicable one

Good to know that doesn't change anything then. I double checked the code to be sure and it's still status.

Issue #202, opened 2022-08-20
Tracker more accurantely tracks enemy moves and the level at which it had that move


UTDZac:
Version: 0.6.1

Currently only four moves are tracked for any enemy Pokemon. This can often times be inaccurate in a sense that encountering the same Pokemon at different levels show incomplete move information. Easiest when shown with an example, see blow.

![missing dream eater](https://user-images.githubusercontent.com/4258818/185767092-7ef7f679-89aa-404b-a5e5-f72b9a5ff71c.png)

Here you can see I've encountered the Forretress quite a few times, mostly around levels 3, 5, 6, and 8. It has four unique moves before level 8, then learns a fifth move at that level. This screen above shows the movelist for the level 8 Forretress, which is wrong because I'm looking at a level 5 Forretress. **It is missing the move Dream Eater (its first level 1 move)**.

My proposed solution is to track ALL unique moves of a Pokemon at all times. It would also associate / update the move with the lowest relevant level of the encounter Pokemon (internal tracked data information). This way the Tracker can more accurate display to the user the proper moves at the proper level, for any level of Pokemon that they run into.

### Example flow of this implementation:
1. User encounters the first Forretress at level 5. They scout all of the moves. The data stored in the tracker would then be:

| Move | Min Level Seen |
| --- | --- |
| Dream Eater | 5 |
| Steel Wing | 5 |
| Psychic | 5 |
| Fire Punch | 5 |

2. The user continues to check the grass and encounters the lv 8. The data stored in the tracker would then be:

| Move | Min Level Seen |
| --- | --- |
| Dream Eater | 5 |
| Steel Wing | 5 |
| Psychic | 5 |
| Fire Punch | 5 |
| Ember | 8 |

3. The user continues checking grass, encountering the lv3 Forretress. Data would be:

| Move | Min Level Seen |
| --- | --- |
| Dream Eater | 3 |
| Steel Wing | 3 |
| Psychic | 3 |
| Fire Punch | 3 |
| Ember | 8 |

### Additional Resulting Feature (new issue to be added later)
Once this is implemented, this allows us to provide the user with more information about any give Pokemon via the Pokemon Info Screen. When viewing the Pokemon info page and looking at "moves learned at a level" they can then click that area to see the full list of knowledge about the moves. This is basically the same info they would have if they were manually tracking the levels of moves for each encounter and updating them along the way (similar to the data tables above).

UTDZac:
[...]

PR #468, opened 2024-05-04
Include known range for learned moves & display definitely known moves

jwunderl:
Run the other day had an instance where gym leader had the same mon as the player. Player ended up speeding through looking at the tracked moves & forgot for a moment they were the last moves the player mon had learned, not the ones you'd see at there level.

The three main changes intended:
* When learning a move, mark the level it is learned as the low, and the last level we know it'll be had as the high (this is just from looking at the learn levels and looking 4 moves ahead)
* When inspecting an opposing mon, make sure moves that the player would know it will have are listed first -- we know it will have the move if minLv is below / equal to it and maxLv is above / equal to it, as any two of the same mon at the same level will have the same move.
* When a move is not guaranteed to be had by the given mon, mark it with an asterisk so the player knows to try and confirm more details in the move history window.

There's definitely some clean up I listed inline, just making a pr to point to as a reference for discord :)

* Handling for mons that learn 2 moves at the same level (could hack this sort of thing in but figured easier to just leave in case there was an obvious approach to it)
* just made a quick concat for the * next to 'not sure if known moves', would want to match the blue color probably
* Could be considered to be doing too much calc for the player: could either add a flag on the tracked move that it was learned on a level up / can exclude from list of moves, or recalculate from the levels, to narrow down displayed moves a bit more when you know the move isn't learned until later
* need to update move button https://github.com/besteon/Ironmon-Tracker/blob/dev/ironmon_tracker/Input.lua#L261, probably better approach is to move most of the datahelper changes to tracker.getMoves & have that take an optional parameter of level.

Couple quick screenshots of what this does after:
* catching level 3 hariyama
* leveling to 19
* returning and seeing a level 3 in grass

image
image

### DIFF
--- a/ironmon_tracker/Program.lua
+++ b/ironmon_tracker/Program.lua
@@ -430,7 +430,24 @@ function Program.update()
 			-- Check if a Pokemon in the player's party is learning a move, if so track it
 			local learnedInfoTable = Program.getLearnedMoveInfoTable()
 			if learnedInfoTable.pokemonID ~= nil then
+				-- start level
 				Tracker.TrackMove(learnedInfoTable.pokemonID, learnedInfoTable.moveId, learnedInfoTable.level)
+				-- track max possible level
+				local data = DataHelper.buildPokemonInfoDisplay(learnedInfoTable.pokemonID)
+				local totalLearnableMoves = #data.p.movelvls
+				-- TODO: this works in most scenarios, but need to figure out handling 2 other things: mons that learn multiple
+				-- 		moves at the same level, and mons
+				for i, moveLvl in ipairs(data.p.movelvls) do
+					if moveLvl == learnedInfoTable.level then
+						-- Tracker.TrackMove(learnedInfoTable.pokemonID, learnedInfoTable.moveId, i)
+						if totalLearnableMoves >= i + 4 then
+							Tracker.TrackMove(learnedInfoTable.pokemonID, learnedInfoTable.moveId, data.p.movelvls[i + 4] - 1)
+						else
+							Tracker.TrackMove(learnedInfoTable.pokemonID, learnedInfoTable.moveId, 100)
+						end
+						break
+					end
+				end
 			end
 
 			if Options["Display repel usage"] and not Battle.inActiveBattle() then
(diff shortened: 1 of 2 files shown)


jwunderl:
Zac pointed out https://github.com/besteon/Ironmon-Tracker/issues/202, so for reference what this would display for the mons mentioned in that issue:

## Forretress after step 3
assuming only the moves descrived, so the history move showing:

ember min 8 max 8
steel wing min 3 max 5
psychic min 3 max 5
fire punch min 3 max 5
dream eater min 3 max 5 (the first move seen, as shown in screenshot as it's the one not listed)

lvl 3-5 show
* dream eater / steel wing / psychic / fire punch confirmed

lvl 6-7 shows
* ember / fire punch / steel wing / psychic same as now, except marked with * on each

lvl 8 shows
* ember as confirmed
* fire punch, steel wing, psychic marked with * on each 

lvl 9+ shows 
* ember, fire punch, steel wing, psychic mark marked with * on each

## krabby
https://github.com/besteon/Ironmon-Tracker/issues/202#issuecomment-1359200969
I'm assuming move order is that they were seen top to bottom to illustrate, but it only swaps out what the *'d ones will be, which would show in the same priority order as they do now

lvl 3-7
* outrage confirmed
* synthesis, grudge, aeroblast have *

lvl 8
* outrage, hydro pump confirmed
* synthesis, grudge *

lvl 11
* charm confirmed
* synthesis, grudge, aeroblast have *

lv 15-17
* spark confirmed
* synthesis, grudge, aeroblast have *

lv 21-22
* return confirmed
* synthesis, grudge, aeroblast have *

lvl 27
* aeroblast and baton pass confirmed
* synthesis, grudge have *

lv 30 
* synthesis, grudge confirmed
* aeroblast, baton pass have *

all other levels get same as now except * on moves (besides whatever is on page 2)

## Altaria
https://github.com/besteon/Ironmon-Tracker/issues/202#issuecomment-1363686571

lvl 27
* pursuit, kinesis, magic coat, extrasensory all confirmed

lvl 28-31
* kinesis, magic coat confirmed
* extrasensory, pursuit marked with *

lvl 32-33
* sandstorm, egg bomb, kinesis, magic coat all confirmed

all other levels get current behavior

In theory we could calc out some more for altaria using the learn levels, but that would be a mistake both for code complexity reasons & doing the work for the player -- just give them the obvious ones and indicate they should treat the others with suspicion / investigate move history to see what they can figure out

PR #472, opened 2024-09-12
Catch Rate Display

jtigues:
- Adding stat offset for catch rate
- Adding logic to save catchRate while loading pokemon data and read it when fighting and viewing wild pokemon


UTDZac:
Looks good. Thanks for this awesome feature

PR #474, opened 2024-09-13
Catch rate feature

jtigues:
- Made the catch rate formula intentionally more inaccurate mathwise so that it is more accurately aligned with the results the game itself arrives at.


UTDZac (commented):
Looks great. I don't have anything to do between now and release, so i'll leave this PR up in the event @Fellshadow also wants to take a quick look before approving the release PR

Fellshadow (approved):
also looks good to me

PR #475, opened 2024-09-19
Convert some numbers to constants

CyanSMP64:
Convert some numbers to constants to support extended nicknames for the upcoming Nat. Dex


Fellshadow (approved):
LGTM, nice bit of cleanup even if there weren't an extension usecase requiring these to be variables

PR #477, opened 2024-09-20
8.6.2 Patch Release

UTDZac:
_(Critical hotfix patch)_

Fixes a critical issue that is current preventing many people from updating their Tracker. No idea how this issue got introduced, happened sometime in the last 4-6 months.

This fix will resolve itself automatically for the broken Trackers when they try to update to this new version. The updater is structured in such a way to first download the new files, then load in any changes to the Update script, and use those to perform the update. In other words, they'll get the fixed file before updating over all other files.

### Release Notes
- Fixed a critical error that prevented the Tracker from self-updating in some cases
- [MGBA] Added the UNFOUGHT() command included in the previous patch

### DIFF
--- a/UpdateOrInstall.lua
+++ b/UpdateOrInstall.lua
@@ -356,7 +356,9 @@ function UpdateOrInstall.buildCopyFilesCommand(extractedFolder, isOnWindows)
 			string.format('echo %s', messages.filesready),
 			string.format('cd "%s"', IronmonTracker.workingDir), -- required for mGBA on Windows
 			string.format('echo %s', messages.updating),
-			string.format('xcopy "%s" /s /y /q', extractedFolder),
+			-- /s: for subdirectories, /y: no overwrite prompts, /q: no msg display, /c: skip files with errors
+			string.format('xcopy "%s" /s /y /q /c', extractedFolder),
+			-- /s: deletes directory tree, /q: no confirmation prompts
 			string.format('rmdir "%s" /s /q', extractedFolder),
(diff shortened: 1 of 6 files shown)


UpdateOrInstall.lua
```diff
@@ -356,7 +356,9 @@ function UpdateOrInstall.buildCopyFilesCommand(extractedFolder, isOnWindows)
 			string.format('echo %s', messages.filesready),
 			string.format('cd "%s"', IronmonTracker.workingDir), -- required for mGBA on Windows
 			string.format('echo %s', messages.updating),
-			string.format('xcopy "%s" /s /y /q', extractedFolder),
+			-- /s: for subdirectories, /y: no overwrite prompts, /q: no msg display, /c: skip files with errors
+			string.format('xcopy "%s" /s /y /q /c', extractedFolder),
```
Fellshadow, line 360: mentioned in the discord but also putting here for completeness:

Main concern with this is what if the file that would cause an error and therefore be skipped by the `/c` flag is a file we *want* to update? Skipping would then cause a partial update to happen which could just break things even more depending on what the skipped files are, so how do we know for sure that the skipped files are *only* ones we don't need to update?

Fellshadow (changes requested):
I'm aware this is critical but i can't approve this without the below comment being addressed, because my concern is the potential of breaking things on the user-end even more down the line

Fellshadow (approved):
merging in following discussion in discord

PR #478, opened 2024-09-20
Tracking Fix

jtigues:
- Fixing bug where on normal level up, the last known level up move learning information was being applied to the current mon. The very last address is called for normal level ups, not just learning a move.
- Also adding a log for the future in case it errors again


UTDZac (approved):
Thanks for investigating and testing this fix. Looks good to me.

PR #479, opened 2024-10-06
New Controller Bindings and UI Screen

UTDZac:
Really happy to finally have this controller screen added in. Let me know if I missed anything obvious. Feel free to give it a test run too for FireRed and/or Emerald.

### Summary of changes
- Add a new UI screen for displaying the current Controller bindings (finally!)
- Add a sub-screen for binding controllers using the controller, no longer have to type them in.
- Tracker now writes a single byte to change a game option for Button Mode to "LR", per the new setting below
- Add new setting for "Override Button Mode to LR", on by default.
- Add new hidden setting for "AlertNewOptionLR", to notify user exactly once of these changes to L/R

### New Screen for Setup > Controls
image

### New Screen for binding a controller button to Tracker function
image


Fellshadow (approved):
LGTM, adding a commit to fix a typo and make a small change as per my comment below

ironmon_tracker/Program.lua
```diff
@@ -1165,6 +1188,16 @@ function Program.isInStartMenu()
 	return startMenuWindowId == 1
 end
 
+---Forcibly change the in-game option for "Button Mode" from "HELP" to "LR"; allowing additional Tracker controls
+function Program.changeGameSettingForLR()
+	local addr2 = Utils.getSaveBlock2Addr()
+	local currentSetting = Memory.readbyte(addr2 + Program.Addresses.offsetOptionsButtonMode)
+	-- 0:NORMAL(HELP), 1:LR, 2:L_EQUALS_A
+	if currentSetting == 0 then
```
Fellshadow, line 1196: Actually, should this *also* check for if button mode is L=A? Since the setting is worded as "Override Button Mode to LR" it'd make sense to me to also change from L=A

Functionally doesn't change a huge amount, but if it's on L=A then there's a potential annoyance if the player has both the route/trainer info open and an in-game menu open and wants to use L to cycle to the previous page

ironmon_tracker/Program.lua
```diff
@@ -1165,6 +1187,16 @@ function Program.isInStartMenu()
...
+---Forcibly change the in-game option for "Button Mode" from "HELP" to "LR"; allowing additional Tracker controls
+function Program.changeGameSettingForLR()
+	local addr2 = Utils.getSaveBlock2Addr()
+	local currentSetting = Memory.readbyte(addr2 + 0x13)
+	-- 0:NORMAL(HELP), 1:LR, 2:L_EQUALS_A
+	if currentSetting == 0 then
+		Memory.writebyte(addr2 + 0x13, 1)
+	end
+end
+
```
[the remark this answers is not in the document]
UTDZac: Good call.

ironmon_tracker/Program.lua
UTDZac: I dont think so. If people intentionally change their setting cause they like using L as A, then i want to leave it like that. The important thing is R being the access button for the other helpful stuffs.

We'll find out in time if people run into an issue with this, and if so we can change it.

PR #480, opened 2024-10-06
8.7.0 Release

UTDZac:
### Release Notes
- You can now press the **R button** outside of battle to see a list of Trainers on the route, or to see early game / Safari Zone pivots. Try it out!!
   - (Press **L+R** together to go back to the normal Tracker screen)
- While battling a Trainer, click the battle balls icon to see their Trainer Card: displays their Pokémon levels, any IVs, AI Script, and more
- New "Edit Controls" screen under Setup > Controls
- On any screen with the left/right paging arrows, press the **L or R** controller button to show previous/next page
- Hidden Power's actual typing and power is now shown for your Pokémon on the Log Viewer after a game over
- Fixed an issue where your Pokémon's Move History would log an incorrect move under rare circumstances
- [Stream Connect] The !weak command now also lists the Pokémon's resistances
- [Stream Connect] The !pivots command now contextually displays highest-level Safari Zone encounters per area
- [Stream Connect] The !ballqueue startup message now shows when you enter the lab instead of immediately
- For Tracker devs: added several new functions in the TrackerAPI file

### Trainer Card Info Screen
image


UTDZac:
Snuck in two additional small changes:
- Add two get Trainer game data functions to `TrackerAPI`
- Pressing both the previous & next page buttons (L+R) at once will perform a screen's "Go Back", for convenience

The back thing is helpful as a quality of life sort of thing. While zooming around and only wanting to hold your controller (not reach over to your mouse), you want to be able to close the info screen that gets pulled up for the route. The most obvious solution here is to use the existing L+R bindings for previous/next page. No new bindings or settings are required to make this work.

ironmon_tracker/Debug/Tracker-StreamerBot.cs
```diff
@@ -534,7 +534,15 @@ private void VerifyOrCreateDataFiles()
...
+				_outboundFile = Path.Combine(directoryOverride, OUTBOUND_FILENAME); // Requests (outgoing)
+			}
+			else
+			{
+				_inboundFile = Path.Combine(dataDirectory, DATA_FOLDER, INBOUND_FILENAME); // Responses (incoming)
+				_outboundFile = Path.Combine(dataDirectory, DATA_FOLDER, OUTBOUND_FILENAME); // Requests (outgoing)
+			}
 
 			// Create required inbound/outbound files
 			_inboundFile = Path.Combine(dataDirectory, DATA_FOLDER, INBOUND_FILENAME); // Responses (incoming)
```
Fellshadow, line 548: the `if` block here gets entirely overridden by the subsequent setting of `_inboundFile` and `_outboundFile` on lines 548,549 below it

ironmon_tracker/Input.lua
```diff
@@ -180,6 +231,42 @@ function Input.checkJoypadInput()
 	end
 end
 
+function Input.infoShortcutPressed()
+	-- At current, don't activate if inside a battle or in menus (Select in-game reorders stuff) or using MGBA
+	if Battle.inActiveBattle() or Program.isInStartMenu() or not Main.IsOnBizhawk() then
+		return
+	end
```
Fellshadow, line 238: is the in-battle check needed now that it's R and not SELECT? (L & R iirc do swap tabs while in the bag so may still want the in-start-menu check)

ironmon_tracker/MGBA.lua
```diff
@@ -538,18 +538,22 @@ MGBA.OptionMap = {
 		getText = function() return Resources.MGBA.OptionDevBranchUpdates end,
 	},
 	[11] = {
+		optionKey = "Override Button Mode to LR",
+		getText = function() return Resources.MGBA.OptionOverrideButtonModeLR end,
+	},
```
Fellshadow, line 543: Does mGBA need this option, and by extension should we be overriding the button mode on mGBA, considering the new shortcut does nothing on mGBA?

Fellshadow (changes requested):
Almost seems fine, though there's one thing that needs fixed (and just a couple other comments i wanted to ask about)

Also there's a typo on one of your release notes:
"Fixed an issue where **`the your`** Pokémon's Move History would log an incorrect move under rare circumstances"
-> should it be "...where a Pokemon's..." maybe?

ironmon_tracker/Debug/Tracker-StreamerBot.cs
UTDZac: Good catch. I forgot to remove those lines.

ironmon_tracker/MGBA.lua
UTDZac: Technically no, but since its here, im going to leave it. If anything, i find this "override to change the default setting from HELP to something useful" to be useful.

ironmon_tracker/Input.lua
UTDZac: Good call, I think i can remove the in-battle check and the in-start menu check. To my knowledge, the L and R buttons have no impact there within the game itself.

I will add another condition to display a more contextual info screen in that case.

Fellshadow: Just double-checked and L and R function like left/right directional buttons while in menus like the bag or on the pokemon summary screen so probably want to leave the in-start-menu check in there

in-battle they don't directly do something unless you go into the same menus (bag, pokemon, summary screen) it seems, unless there's a doubles functionality or something

Fellshadow (approved):

PR #483, opened 2024-10-11
8.7.2 Patch Release

UTDZac:
### Release Notes
- Fixed an issue where pressing the R button to mark a stat would incorrectly show a different screen
- Fixed an issue with Nat Dex compatibility for the new LR Button Mode Tracker setting override
- [FRLG] Fixed an issue where the Rival's true name was revealed instead of the name entered by the player
- Added a way to unbind a Tracker Controller button
- The Trainers on route info screen now shows all trainers for a given dungeon/area if more than one floor.
- [Stream Connect] Can now use the parameter "nodoubles" with the !unfought command to exclude double battles
- Several minor UI improvements and internal data fixes


Fellshadow (approved):
LGTM - don't forget to update the stream connect wiki too with the command update here

PR #484, opened 2024-10-19
Tracker Notebook: Pokémon and Trainer info screens

UTDZac:
This feature adds a long awaited way of viewing/editing Tracker notes as well as other game information for the game in progress. The **Notebook** button is located on the Tracker's main settings page. I swapped around the Theme and Language button locations based on usability and least impact.

Feel free to give this a test run, it's pretty fun to use. Lots of little pieces of information. If you have any thoughts on ways I can improve any of the screens or utilize the empty space, let me know. I don't mind the empty space for easier viewing, as the important information will stand out more.

## Screen: Tracker Notebook
image

From here, the player can choose to review all the notes about Pokémon they've encountered thus far, or otherwise those with manually entered notes, by clicking the **"Pokémon Seen"** button. "Notes" including anything tracked about the Pokémon: stats, ability, encounter, level, moves, or a written note by the player. The player can also review all of the trainers they have battled thus far, in a very loose weakest-to-hardest ordering by route area, via the **"Trainers Fought"** button. This is the same output and ordering for the `!unfought` [Stream Connect command](https://github.com/besteon/Ironmon-Tracker/wiki/Stream-Connect-Guide#list-of-commands).

> _[NOT IMPLEMENTED]_ I've considered adding a third button at the bottom of this screen for **"Search"**, to allow searching for specific abilities or moves and listing all the Pokémon with that info. Wasn't sure the best way to implement this, as it would require an on-screen keyboard like the log viewer search or an annoying pop-up window which I don't want to do.

## Screen: All Pokémon Seen
image

By default, this page lists all the Pokémon the player has seen thus far. They can filter by letter, which is simply the starting letter of a Pokémon name; not much space to do anything else here. The **"All"** checkbox will list all Pokémon in the entire game, to allow the player to edit notes for any Pokémon and not just those they've encountered.

The **"Seen: #"** is a total count of all encounters, both wild and trainer. If both types have been encountered, I have it display as **"Seen: 2(T) + 8(W)"** as seen twice on trainers and 8 times in the wild. Unsure of a better way to display this in the small space.

The occasional `+`/`-`/`=` symbols in the corner is a simply summary of the markings noted for the Pokémon by the player. If there are more `+` markings among the stats, it will display a `+`. This was a quick way to easily identify which Pokémon are likely threats while scanning through the list here.

## Screen: Pokémon Note View
image

This screen is very similar to the original Tracker Screen when viewing an opposing Pokémon in battle. It affords all the same usability, info lookup by clicking on anything, stat marking, note taking, etc.

Some noticeable differences are:
- Empty space below the Pokémon's name; no "current level" to display because not in battle
- No encounter route name, instead shows counts for trainers and wilds individually
- Less info about an individual move but more moves listed (up to 8)

## Screen: All Trainers by Area
image

This screen lists all route areas/dungeons that have trainers in them that haven't been fought yet. By default, it does *not* **"Show completed"**, I simply checked it here for the screenshot. When playing FRLG, there is also a **"Sevii"** checkbox to show those locations as well (Off by default). Clicking on any given route/dungeon will then display all the trainers there, using the existing `TrainersOnRouteScreen`.

The algorithm used to determine the what routes to list and the order to list them is as follows:
- Order all trainers from weakest-to-strongest based on the average level of the Pokémon on their team
- Starting with the first trainer, if not defeated, add it's route/dungeon to the next spot on the list
   - Then check the other trainers on its same route/dungeon
- Repeat this process until all trainers have been checked

The upside to this algorithm is that it is game-agnostic and doesn't imply any sort of routing strategy or pathing. It is simply just weakest-to-strongest average level of a team. The downside to this is that sometimes the ordering will feel weird, such as displaying Route 4 before Pewter Gym when playing FRLG.


Fellshadow (approved):
LGTM, just one small comment about commented out lines in case you had forgotten about them or something

UTDZac:
Good catch. I intentionally left this commented out in the event I change
up the UI of the screen to fit in this information.

At current, I like the simplicity and compactness of the initial page with
the one value that is there

PR #486, opened 2024-10-25
New Feature: Stat marking grading score sheet

UTDZac:
This feature adds a fun little note score sheet (graded stat markings) after a run ends, accessible on the Game Over screen. It checks all **non-speed** markings the player noted for Pokémon they fought against, and compares their markings to the actual base stats of that Pokémon.

Great markings are those within the acceptable range for that stat marking. Poor markings are outside that range. There is also a margin of error to take into account the uncertainty of markings.

A letter grade will also be displayed in the upper-right corner. This grade can be A, B, C, or D. If all stat markings were graded correctly, then it will be an A+ to indicate a perfect score.

> Note: While these grading ranges for "correct stat markings" are subjective, I still want to do something like this as it seems both fun to look at and encouraging to take better notes. I used a few different samples of data to test and confirm the grading system is within reason, likely compatible with most people's methods of stat marking.

### Stat Grade Ranges
| Symbol | Stat Marking | Range |
| --- | -- | -- |
| `[+]` | High | 115 - 255 |
| `[-]` | Low | 11 - 70 |
| `[=]` | Medium/Equal| 70 - 115 |

All stat ranges are inclusive. The margin of error is +/- 25 for either side of the range.

For example, if the ATK stat with a `[+]` and that Pokémon's base stat value is 100, then it will be graded as correct, or "Great mark", as its within the acceptable range and margin of error.

### Stat Marking Score Sheet screen
image

### DIFF
--- a/ironmon_tracker/screens/StatMarkingScoreSheet.lua
+++ b/ironmon_tracker/screens/StatMarkingScoreSheet.lua
@@ -1,0 +1,477 @@
+StatMarkingScoreSheet = {
+	Colors = {
+		text = "Default text",
+		highlight = "Intermediate text",
+		positive = "Positive text",
+		negative = "Negative text",
+		border = "Upper box border",
+		boxFill = "Upper box background",
+	},
+	Data = {},
+}
+local SCREEN = StatMarkingScoreSheet
+
+SCREEN.StatMarkingRanges = {
+	MarginOfError = 25,
+	[Constants.STAT_STATES[1].text] = { min = 115, max = 255, }, -- [+] positive
+	[Constants.STAT_STATES[2].text] = { min = 11, max = 70, }, -- [-] negative
+	[Constants.STAT_STATES[3].text] = { min = 70, max = 115, }, -- [=] equal
+}
+
+SCREEN.PixelImages = {
+	SMALL_X = {
+		{1,0,0,0,1},
+		{0,1,0,1,0},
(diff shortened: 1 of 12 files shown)


ironmon_tracker/screens/StatMarkingScoreSheet.lua
```diff
@@ -0,0 +1,630 @@
...
+
+---Returns a pixel-image table that represents a letter grade from A to D.
+---@param percentile number
+---@return table pixelImage
+function StatMarkingScoreSheet.getGradePixelImage(percentile)
+	percentile = percentile or 0
+	if percentile == 100 then
+		return SCREEN.PixelImages.GRADE_A
+	elseif percentile >= 90 then
+		return SCREEN.PixelImages.GRADE_A
```
Fellshadow, line 541: these two if blocks are returning the same thing, any reason they're separated?

Fellshadow (approved):
LGTM, just one small comment

ironmon_tracker/screens/StatMarkingScoreSheet.lua
UTDZac: I had originally thought I would have two pixel-images, one for "A+" and one for "A". I'll still leave this as-is, since the performance impact is nonexistent, and I may want to change the symbol in the future.

PR #489, opened 2024-11-11
fix map layout id of underwater maps

CyanSMP64:
fixes a mistake on my part from #439 
back when i wrote it, i falsely assumed the route data entry numbers were map ids instead of layout ids

### DIFF
--- a/ironmon_tracker/data/RouteData.lua
+++ b/ironmon_tracker/data/RouteData.lua
@@ -4064,7 +4064,7 @@ function RouteData.setupRouteInfoAsRSE()
 			{ pokemonID = 319, rate = 0.40, minLv = 30, maxLv = 35, },
 		},
 	}
-	RouteData.Info[51] = {
+	RouteData.Info[274] = {
 		name = "Route 124 Water",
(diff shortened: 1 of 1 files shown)


UTDZac:
I'm not sure how to understand what the issue is here. Working with Emerald is difficult for me, as I do not play it enough to know what is or isn't correct.

From my understanding, we had the correct route data already.

Can you share some information to explain why its incorrect? Help me understand why this change is required?

CyanSMP64:
at the time i wrote #439, i wasn't made aware of the fact that each map in the gen 3 games uses a layout id in their headers; many maps share layouts, such as generic houses, pokemon centers, and poke marts. looking at the route data now i realise that each route entry uses the layout id number (https://github.com/CyanSMP64/pokeemerald/blob/master/include/constants/layouts.h)

the underwater maps for routes 124, 126, 127, 128 follow a sequential order in the map groups (https://github.com/CyanSMP64/pokeemerald/blob/master/include/constants/map_groups.h) whereas the route 124 underwater layout has id number 274 while the others have 51-53.

this was correct until #439 was merged

PR #494, opened 2024-11-18
Code to track initial moveset of Pokemon used

UTDZac:
This solves an outstanding issue where the initial 4 moves of a player's lab/pivot Pokemon are never tracked, since the conditions for tracking a move were "used in battle" or "learned while leveling up".

This remedies that issue by doing a 1-time tracking of the current moveset when battling a trainer (player decided to run this Pokemon for real). It also only tracks moves within its initial moveset if that Pokemon could have naturally learned that move by level up.

### Before 1st Trainer Battle
image

### After 1st Trainer Battle
image

### DIFF
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -445,6 +445,16 @@ function Battle.updateTrackedInfo()
 		end
 	end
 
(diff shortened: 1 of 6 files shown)


jtigues (approved):
Looks good to me. Noting that there is a minor chance for unknown information being revealed for your second pokemon in double battles, but for the most part won't come up, and won't matter.

Fellshadow (approved):

PR #499, opened 2024-12-20
New Feature: Game Profiles for New Runs

### LINKED ISSUE #450: [Feature Request]: Add the ability to save tracker settings as presets.
### Tell us about your request

For example --
  
Say I'm playing NatDex FireRed, and I wanna switch to vanilla FireRed. I have to reset the attempts counter, point the tracker to the correct version of the randomizer, the ROM I wanna use, and the settings file while navigating to each individual element, which are in three different places in my directory. Maybe I might have to change the welcome message and my favorites, or maybe I like to have a different tracker theme depending on what I'm playing. AND on top of everything, I have to disable the NatDex extension so that the tracker can read the vanilla FireRed information correctly. 

The point is, switching games is a lot of work on the user's end, so being able to simplify it would help a ton. Hence, why I'd suggest adding a little button somewhere where you can load/save presets so that all of what I just described can be done in just a couple clicks. 

### Additional Information

_No response_

UTDZac:
Closes #450 
 
This feature adds the ability to create, manage, and use different New Run setup profiles for different games or rulesets. This has been a highly requested feature, especially for people who play different Ironmon rulesets all the time.

In short, it works similar to how New Run setup worked in the past, but all of the settings are saved as a Profile object. All profiles are stored within a `NewRunProfiles.json`, located in the `ironmon_tracker` folder.

In addition to storing the New Run settings as a profile, this PR also changes how Tracker Notes Data (`.TDAT` files) are saved. Previously, it was saved as a single file for a single game version (such as Fire Red). Now, it saves a file for each profile, allowing for multiple "Fire Red" data files if more than one profile is using the Fire Red game version. These files are stored in the `tracker_notes` folder (new as of this PR).

### Example: List of Profiles
image

### Example: Active Profile
image

### Example: Adding or Editing a Profile
image


Fellshadow (approved):
looks good to me

PR #500, opened 2024-12-22
Override moves shown to use known moves if able

### LINKED ISSUE #498: [Bug]: Tracker fails to display all 4 moves used by enemy Pokémon in battle
### Describe the bug

This is a follow-up to this issue: https://github.com/besteon/Ironmon-Tracker/issues/202

Just reading through it again has given me an idea for how to solve the more common case here: where all 4 moves do get used in battle and those should be the 4 moves displayed.

The solution is to just track the moves used for the Pokémon (plural) in the current battle, like we track many battle related things. If all 4 moves are known for a given Pokémon, the Tracker just displays those; otherwise is falls back to using the original implementation.

Seems easy enough to code while also avoiding the legacy code logic I've been really trying to avoid changing

## Example Scenario
image image

In this case, the lv.32 Shedinja used moves: Spikes, Sweet Scent, Aromatherapy, and Torment, but the Tracker is displaying a non-helpful old move.

v8.9.0

Windows

Bizhawk 2.9

### Relevant log output

_No response_

### Additional Information

_No response_

UTDZac:
Closes #498 

In some cases, while battling and viewing an enemy Pokémon, that Pokémon will use all four of its moves. When it does, show those four moves on the tracker screen instead of the loose ordering of otherwise "tracked moves" it used to show.

This fixes the bug where 1 of the 4 moves wouldn't be shown even if all were known during that single battle.

### DIFF
--- a/ironmon_tracker/Battle.lua
+++ b/ironmon_tracker/Battle.lua
@@ -398,6 +398,7 @@ function Battle.updateTrackedInfo()
 										local battlerMon = Tracker.getPokemon(battlerTransformData.slot,battlerTransformData.isOwn)
 										if battlerMon ~= nil then
 											Tracker.TrackMove(battlerMon.pokemonID, lastMoveByBattler, battlerMon.level)
+											Tracker.recordBattleMoveByPokemonLevel(battlerMon.pokemonID, lastMoveByBattler, battlerMon.level)
 										end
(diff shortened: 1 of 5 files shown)


Fellshadow (approved):
Seems good to me

Issue #497, opened 2024-12-16
[Feature Request]: Store GameSettings.lua addresses as a JSON


UTDZac:
## Goal of this change
Makes it easier for custom ROM hack developers to conveniently provide a list of changed addresses for the Tracker to load up and use.

## Process of implementation
Move the hard-coded addresses from `GameSettings.lua` to JSON files, one file for each supported game version (such as Emerald, FireRed v1.0, FireRed v1.1, etc.)

## Example custom code extension
Here is a snippet of code showing how a custom code extension uses this concept to replace the used addresses:
```
self.offsetFiles = {
    [2] = "offsets_em.json",
    [3] = "offsets_fr.json"
}
self.extensionKey = "MAXExtension"
local offsetFile = self.offsetFiles[GS.game] or "offsets.json"
local filePath = FileManager.getCustomFolderPath() .. self.extensionKey .. FileManager.slash .. offsetFile
local off = FileManager.decodeJsonFile(filePath)
--pass the off table into a function that loads everything into the game settings
```

## Example JSON file
Definitely want a bit more structure than this, but here is a loose example
```
{"gBattleTerrain":33697776,"gBattleStruct":33703068, ... }
```

UTDZac:
Completed in v9.0.0

PR #503, opened 2024-12-26
add constants for battle details screen

CyanSMP64:


UTDZac:
I am in the process of massively overhauling the BattleDetailsScreen, and intend to automate the process for updating these constants with a rework of GameSettings.lua. The goal is to automatically load in all of these changes via your own extension, without the need of any changes to the core tracker code.

https://github.com/besteon/Ironmon-Tracker/issues/497

This is taking me some time to implement, so i'll need a bit of patience here.

UTDZac:
As an added note, the `dev` branch's BattleDetailsScreen will likely be non-function for the next few days. I am greatly reworking the code to fit it more inline with a typical tracker screen

UTDZac:
Thanks again for looking into this. I believe this PR is resolved per new release v9.0.0. These values are exposed as part of the BattleDetailsScreen object, and easy to override with the new json support added for GameSettings addresses. Let me know if you have any questions on using it, or if I missed any addresses.

PR #515, opened 2025-03-09
feat: Use Bizhawk rom path as a fallback for logpath

AndreVolpi:
1. Remove all New Run profiles from the Tracker
2. Create a new ironmon rom named "TestRom Roguemon.gba"
3. Have the associated log file named "TestRom Roguemon.gba.log" also created, and present in the same folder
4. Open the rom from step 2 into bizhawk, and open the tracker
5. Go to Extras -> View Log -> Open Log

Desired result: Open the log from step 3
Current result: A file prompt popup appears instead asking for the location of the log file


UTDZac (approved):
Looks good to me. A simple change to check for an obvious potential file path of the log file, using that if found.

PR #519, opened 2025-03-21
Separate variable power logic per individual move

### LINKED ISSUE #518: [Feature Request]: Make the variable move power calculation more generic, allowing extensions to use/add to it
### Tell us about your request

Currently the calculation for variable power moves in [Datahelper.lua](https://github.com/besteon/Ironmon-Tracker/blob/a3ffaf9e80318bf1193f6bdc7a1e38002f089a16/ironmon_tracker/data/DataHelper.lua#L297-L328) use hardcoded move IDs to determine what moves to display variable powers for, and how to calculate the variable power

This means extensions such as the MAX extension cannot have newly added moves such as Grass Knot (which functions in the same way as low kick) display a variable power when the user has the option enabled

We should make this more generic, probably by moving the move IDs to an ENUM or list, that is exposed to extensions so that they can add in their own move IDs for moves they want variable power displayed for

### Additional Information

_No response_

UTDZac:
Closes #518 

Move the logic for calculating variable power info about a move from the DataHelper into separate, individual functions defined in MoveData. In this way, developers can adjust these formulas if needed and/or add new formulas for other/new moves.

### DIFF
--- a/ironmon_tracker/data/MoveData.lua
+++ b/ironmon_tracker/data/MoveData.lua
@@ -1,7 +1,10 @@
 MoveData = {}
 
 MoveData.Values = {
+	GuillotineId = 12,
+	HornDrillId = 32,
 	LowKickId = 67,
+	FissureId = 90,
 	FlailId = 175,
 	ReversalId = 179,
 	ReturnId = 216,
@@ -12,6 +15,7 @@ MoveData.Values = {
 	SkillSwapId = 285,
 	WeatherBallId = 311,
 	WaterSpoutId = 323,
+	SheerColdId = 329,
 
 	-- The below are used by the BattleDetailsScreen
 	PayDayId = 6,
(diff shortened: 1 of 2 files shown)


ironmon_tracker/data/MoveData.lua
```diff
@@ -352,6 +439,20 @@ function MoveData.calcHiddenPowerTypeAndPower(ivs)
 	return moveType, movePower
 end
 
+---Adjusts the move table data based on any variable damage calculations, or other attributes; No return, as this edits the move table directly.
+---@param move table
+---@param sourcePokemon? table Optional, as not all move adjustment calculations require a source and/or a target
+---@param targetPokemon? table Optional, as not all move adjustment calculations require a source and/or a target
+function MoveData.adjustForVariablePower(move, sourcePokemon, targetPokemon)
```
Fellshadow, line 446: Same as above, these don't just adjust move power so this should be more generic e.g. `adjustVariableMoveValues` or something

ironmon_tracker/data/MoveData.lua
```diff
@@ -135,6 +135,93 @@ MoveData.TypeToEffectiveness = {
 	steel = { fire = 0.5, water = 0.5, ice = 2, rock = 2, steel = 0.5, electric = 0.5 },
 }
 
+-- Individual formulas for calculating any changes to a move based on contextual information, such as being in battle
+MoveData.VariablePowerAdjustmentFuncs = {
+	[MoveData.Values.WeatherBallId] = function(move, sourcePokemon, targetPokemon)
+		if not Battle.inActiveBattle() then return end
```
Fellshadow, line 141: this return is redundant as there's already an early return in `Utils.calculateWeatherBall`, though arguably the latter one in the Utils func is made redundant by this one and could be removed

No harm keeping both of course, just mentioning as a small bit of clean-up that could be done

Fellshadow (commented):
Sorry it took a while to get to looking at this, i'm fine with it as-is but left some comments on small things that could be cleaned up or made more consistent

Fellshadow (approved):

PR #526, opened 2025-04-16
Patch 9.2.3

UTDZac:
### Release Notes for 9.2.3
- [9.2.3] Fixed an issue where the Fairy type wasn't displaying for Nat. Dex coverage calculator
- [9.2.3] Added a [GachaMon Help Wiki](https://github.com/besteon/Ironmon-Tracker/wiki/GachaMon-Collectable-Card-Game) link for learning more about it
- ~~[9.2.3] Fixed an issue where Stars wouldn't show on the Tracker if Survival PC heal tracking was enabled~~
- [9.2.3] Changed the behavior of conflicting Gachamon stars and PC Heal tracking options so that only one of the two can be toggled ON at any one time

I understand for the third point about pc heals and survival that this isn't technically accurate description, but I wanted to use as few words as possible to describe the outcome of this change, such that people won't get confused now if they disable-then-enable the "Show stars" option for GachaMon. It will now just work, no questions asked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,8 +62,10 @@ Trackers for other games:
 
 ## Latest Changes
 
-### Easier Extension Installs & Updates
+### GachaMon! A Collectable Card Game
 ![The GachaMon Collectable Card Game about page, explaining how the new feature works. Get GachaMon cards by playing the game, catching Pokémon, and defeating trainers. Cards are rated from 1-5 stars.](https://github.com/user-attachments/assets/500037d8-0cf4-4f91-8267-c00a852873ad)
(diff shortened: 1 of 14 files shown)


Fellshadow (approved):
For the third point, i feel the release note could still be better worded as it isn't really a "fix" more a change in behaviour, and people might be confused about why they toggle a setting at one point for it to then be disabled later

Maybe something like
>Changed the behaviour of conflicting Gachamon stars and PC Heal tracking options so that only one of the two can be toggled on at any one time

PR #528, opened 2025-04-28
Allow extensions to define required extensions for use

UTDZac:
Some extensions require other extensions to be loaded/started before they themselves can properly startup. This commit adds support for that by adding a new attribute to extension self object: `requiredExtKeys`

### DIFF
--- a/ironmon_tracker/CustomCode.lua
+++ b/ironmon_tracker/CustomCode.lua
@@ -420,7 +420,9 @@ function CustomCode.downloadAndInstallExtensionFiles(githubRepoUrl, folderNamesT
 	return true
 end
 
--- Simulates an interface-like function execution for custom code files
+---Simulates an interface-like function execution for custom code files
+---@param funcLabel string The name of the function (used for error logging)
+---@param ... any Optional parameters for the function
 function CustomCode.execFunctions(funcLabel, ...)
 	if not CustomCode.isEnabled() then return end
 
@@ -434,6 +436,11 @@ function CustomCode.execFunctions(funcLabel, ...)
 	end
 end
 
+---Executes an extension's function.
+---@param extensionKey string Usually the filename of the extension
+---@param functionLabel string The name of the function (used for error logging)
+---@param functToExec function The function itself
+---@return boolean success
 function CustomCode.tryExecute(extensionKey, functionLabel, functToExec)
 	CustomCode.extBeingExecuted = extensionKey
 	CustomCode.funcBeingExecuted = functionLabel
@@ -443,6 +450,8 @@ function CustomCode.tryExecute(extensionKey, functionLabel, functToExec)
 	return result
 end
 
+---Logs an error message
+---@param err any
 function CustomCode.logError(err)
 	err = tostring(err)
 	local errorMessage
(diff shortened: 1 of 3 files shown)


ironmon_tracker/CustomCode.lua
```diff
@@ -458,6 +467,50 @@ function CustomCode.logError(err)
...
+	-- Keep trying to startup extensions in a cascading fashion, until no new ones get started
+	local anyNewStarted = false
+	local unstartedExtensions = CustomCode.EnabledExtensions
+	for _ = 1, 999, 1 do
+		anyNewStarted, unstartedExtensions = _tryStartupExtensions(unstartedExtensions)
+		-- No new extensions were started, no need to keep trying
+		if not anyNewStarted or #unstartedExtensions == 0 then
+			break
+		end
+	end
```
Fellshadow, line 511: a while loop feels like it'd be more suitable here, aside from the above issue

ironmon_tracker/CustomCode.lua
```diff
@@ -458,6 +467,50 @@ function CustomCode.logError(err)
...
+		local anyNewStarted = false
+		local skippedExtensions = {}
+		for _, ext in pairs(extensionsToStartup or {}) do
+			local allRequiredAreStarted = true
+			for _, requiredKey in pairs(ext.selfObject and ext.selfObject.requiredExtKeys or {}) do
+				if not Utils.isNilOrEmpty(requiredKey) and not startedExtensionKeys[requiredKey] then
+					allRequiredAreStarted = false
+					break
+				end
+			end
```
Fellshadow, line 486: this checks for if an extension has any `requiredExtKeys` defined and if those required keys have already been started, but does not check if those required extensions *exist* to start-up

so say roguemon lists natdex as required, but the natdex extension has not been installed, this block will cause the later extension startup loop to cycle all 999 times as it will keep skipping roguemon due to natdex not being started (natdex cannot be started, it does not exist)

this might not be performance-costly, if all it's doing is skipping the extension, but I feel it should be handled better than "it loops 999 times, doing nothing each time" (and should inform the user so they are aware that the extension isn't starting because of a missing dependancy)

Fellshadow (commented):
In its current state I wouldn't feel comfortable with this in a release, though this is going into dev so i'll leave it up to you if you want to address the issue in the PR or in dev later

ironmon_tracker/CustomCode.lua
UTDZac: i purposely want to avoid a `while` loop whenever possible, since if we accidentally miss on the condition to exit, it will freeze the emulator. thus, a `for` loop with just a large iteration cap is generally safer for emulator

ironmon_tracker/CustomCode.lua
UTDZac: in its current state, it wont loop 999 times if nothing new gets installed. each iteration/cascade check will try and install. if and only if a new thing gets installed, will it loop/iterate again to try and install more stuff. if at any point nothing changes from one iteration to the next, it simply exists

Fellshadow (approved):
I think this looks fine now, thanks for the quick work on it

PR #545, opened 2026-04-13
Add base stat spread to MGBA Tracker Screen display

238989234:
Show individual base stats alongside live stats for own Pokemon on the Battle Tracker display. Each stat row now shows the base stat value before the calculated live stat.

Changes:
- PokemonData.lua: Add baseStats to BlankPokemon
- DataHelper.lua: Expose baseStats in tracker screen data
- MGBADisplay.lua: Render base stats in stat column for own Pokemon


ironmon_tracker/MGBADisplay.lua
```diff
@@ -759,7 +759,12 @@ MGBADisplay.LineBuilder = {
 		local formattedStats = {}
 		for _, statKey in ipairs(Constants.OrderedLists.STATSTAGES) do
 			local statValue = Utils.inlineIf(data.p[statKey] ~= 0, data.p[statKey], Constants.BLANKLINE)
-			formattedStats[statKey] = Utils.formatUTF8("%-5s" .. justify3 .. "%-2s", data.p.labels[statKey], statValue, data.p.stages[statKey])
+			if data.x.viewingOwn and data.p.baseStats then
+				local baseValue = Utils.inlineIf(data.p.baseStats[statKey] ~= 0, data.p.baseStats[statKey], Constants.BLANKLINE)
+				formattedStats[statKey] = Utils.formatUTF8("%-4s" .. justify3 .. " " .. justify3 .. "%-2s", data.p.labels[statKey], baseValue, statValue, data.p.stages[statKey])
```
Fellshadow, line 764: This is something that can be discussed if we go ahead with adding in base stats in some form on the mGBA tracker, but I find the base stats being wedged in-between the stat name and actual stat value a little cluttered / confusing here

Fellshadow (changes requested):
Hi, appreciate the interest in contributing to the tracker but there's a few issues with this PR

Firstly, as per our [Contributions info](https://github.com/besteon/Ironmon-Tracker?tab=readme-ov-file#branches-and-processes), PRs should be targeting our dev branch - i've changed the base here for you but just letting you know

As for the feature itself, we don't want this in its current state. The way this is implemented currently will always show the base stats of your randomized pokemon which is information the player does not have unless they look in the log file (not allowed in ironmon unless you're playing open book). Unless the player is playing a vanilla game or specifically has open book mode enabled these base stats should be hidden

That said, I do see potential value of this feature for those usecases of vanilla / open book mode, though neither are currently exposed on the mGBA tracker so might take a good bit more work to implement

I'll leave this open for now, unless @UTDZac feels differently, but I'd highly encourage discussing this with us in the tracker-dev-chat channel of the [ironmon discord](https://discord.com/invite/jFPYsZAhjX) to decide on a potential implementation here that fits the tracker's values

ironmon_tracker/MGBADisplay.lua
238989234: Thank you for the feedback! Definitely want to respect the rules of ironmon as intended and fully understand that base stat spread should not be included in the ruleset. As you mentioned the intent would be for open book, two concrete example use cases that motivated this PR are:

- looking at the logs to check stat totals of what ended the run
- for vanilla gameplay or non-ironmon randomizers, being able to look at base stat totals if shuffled/randomized.

But happy to continue the discussion and seek input over Discord, will start a thread there. Cheers!

PR #547, opened 2026-06-28
Fix TAR "Child Returned Status 128"

UTDZac:
Fix TAR "Child Returned Status 128"

For some people's computers, they receive a "Cannot connect to X: resolve failed" error, code: "Child returned status 128". The fix, accordingly to some online forums, is to include the `--force-local` in the tar command.

https://unix.stackexchange.com/questions/13377/tar-extraction-depends-on-filename

### DIFF
--- a/UpdateOrInstall.lua
+++ b/UpdateOrInstall.lua
@@ -62,6 +62,7 @@ function UpdateOrInstall.setupEmulatorSpecifics()
 	end
 
 	-- This function doesn't exist in Bizhawk, only mGBA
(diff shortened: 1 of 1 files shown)


Fellshadow (approved):
Seems simple enough and fine to me from what I can tell

UTDZac:
As a heads-up, I've reverted this commit from the dev branch. This "fix" will break everyone else. I'm not sure how best to resolve this issue for people who come across it, other than a case by case basis.

PR #550, opened 2026-08-31
Fix unintentional reveal of certain log data

### LINKED ISSUE #549: [Bug]: Incorrect info from memory reads sometimes displayed when viewing previous logs
### Describe the bug

Observed during ironmonvs

When you have seed A where any given mon has a singular ability, it correctly shows that singular ability when viewing the "current" log file

After creating a new seed, B, where that same mon now has 2 abilities, viewing seed A's log (now the "previous" log) will display the original singular ability from seed A _and_ the new second ability from seed B

This is an issue in the tracker's log parser, as evidenced by example B below

Example A:
Metagross when viewed as current seed has truant, when viewed as previous seed it shows both truant and pure power
Image

Example B (with hard log file confirmation):
Anorith has just Chlorophyll in current seed, when viewed as previous seed it grabs Poison Point from the new "current" seed. Mon was not encountered in the wild
> in my screenshots for the anorith, i died and then clicked view the logs before making the new seed. then made the new seed and viewed previous log without checking the new one
Image
Image

### Steps to Reproduce

1. Randomize a new seed
2. View the current log file, confirm the ability of a mon with 1 ability
3. Randomize a new seed
4. View the previous log file, confirm that same mon from 2 now shows 2 abilities

9.3.1

Windows

Bizhawk

UTDZac:
Incorrect assumptions were made to checking for missing log data and filling it in with data from the current game loaded. This now has an additional check to verify it's okay to view that log data (pretty much only when playing a vanilla game or open book mode)

### DIFF
--- a/ironmon_tracker/data/DataHelper.lua
+++ b/ironmon_tracker/data/DataHelper.lua
@@ -615,14 +615,30 @@ function DataHelper.buildPokemonLogDisplay(pokemonID)
 	data.p.id = pokemonInternal.pokemonID or 0
 	data.p.name = RandomizerLog.getPokemonName(pokemonID)
 	data.p.bst = pokemonInternal.bstCalculated or pokemonInternal.bst or Constants.BLANKLINE
+
 	data.p.types = {
-		pokemonLog.Types[1] or pokemonInternal.types[1],
-		pokemonLog.Types[2] or pokemonInternal.types[2],
+		pokemonLog.Types[1],
+		pokemonLog.Types[2],
 	}
+	-- Sometimes types info isn't available from the log (i.e. vanilla game). As such, show if allowed.
+	if not data.p.types[1] and PokemonData.canShowUnknownTypes() then
+		data.p.types[1] = pokemonInternal.types[1]
+	end
+	if not data.p.types[2] and PokemonData.canShowUnknownTypes() then
+		data.p.types[2] = pokemonInternal.types[2]
+	end
+
 	data.p.abilities = {
-		pokemonLog.Abilities[1] or PokemonData.getAbilityId(pokemonID, 0),
-		pokemonLog.Abilities[2] or PokemonData.getAbilityId(pokemonID, 1),
+		pokemonLog.Abilities[1],
+		pokemonLog.Abilities[2],
 	}
+	-- Sometimes ability info isn't available from the log (i.e. vanilla game). As such, show if allowed.
+	if not data.p.abilities[1] and PokemonData.canShowUnknownAbilities() then
+		data.p.abilities[1] = PokemonData.getAbilityId(pokemonID, 0)
+	end
+	if not data.p.abilities[2] and PokemonData.canShowUnknownAbilities() then
+		data.p.abilities[2] = PokemonData.getAbilityId(pokemonID, 1)
+	end
 
 	-- The following are all Randomizer Log information
 	data.p.helditems = pokemonLog.HeldItems or Constants.BLANKLINE -- unsure how this is formatted
@@ -631,7 +647,7 @@ function DataHelper.buildPokemonLogDisplay(pokemonID)
 	for _, statKey in ipairs(Constants.OrderedLists.STATSTAGES) do
 		if pokemonLog.BaseStats ~= nil then
 			data.p[statKey] = pokemonLog.BaseStats[statKey] or 0
-		elseif pokemonInternal.baseStats ~= nil then
+		elseif pokemonInternal.baseStats ~= nil and PokemonData.canShowUnknownStats() then
 			data.p[statKey] = pokemonInternal.baseStats[statKey] or 0
 		else
 			data.p[statKey] = 0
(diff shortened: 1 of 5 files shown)


Fellshadow:
will have a proper look later when I get the time but one thought that comes to mind immediately:
> This now has an additional check to verify it's okay to view that log data (pretty much only when playing a vanilla game or open book mode)

I think there also needs to be an additional check that it's the _current_ log that is being viewed, otherwise this would produce the same bug when viewing a past seed's log in open book mode

UTDZac:
Good note. I'd rather be safe and find a way to check for previous vs current log, which isn't easily doable right now. I'll have to review the code later to find a way to make this happen; shouldn't be too hard.

AndreVolpi:
In what scenario would it need to load one ability from one source and another from another source? Shouldn't it be if both are empty, then pull from memory, otherwise it is what it is?

ironmon_tracker/data/DataHelper.lua
```diff
@@ -615,14 +615,30 @@ function DataHelper.buildPokemonLogDisplay(pokemonID)
...
+		pokemonLog.Types[2],
 	}
+	-- Sometimes types info isn't available from the log (i.e. vanilla game). As such, show if allowed.
+	if not data.p.types[1] and PokemonData.canShowUnknownTypes() then
+		data.p.types[1] = pokemonInternal.types[1]
+	end
+	if not data.p.types[2] and PokemonData.canShowUnknownTypes() then
+		data.p.types[2] = pokemonInternal.types[2]
+	end
+
```
Fellshadow, line 634: Agree with [Kurumas' comment](https://github.com/besteon/Ironmon-Tracker/pull/550#issuecomment-5496421829) here, I don't see a scenario where it would be useful to pull one type from the log and the other from memory so these should probably be combined and the result is one of:
1. Both types pulled from log file, because they are present
2. Both types pulled from memory, because neither are present in the logs and the scenario is ok to read from memory
3. No types shown because neither of the above apply (which is probably fine tbh, just indicates something wrong with the log that the user might want to resolve)

ironmon_tracker/data/DataHelper.lua
```diff
@@ -615,14 +615,30 @@ function DataHelper.buildPokemonLogDisplay(pokemonID)
...
+		pokemonLog.Abilities[1],
+		pokemonLog.Abilities[2],
 	}
+	-- Sometimes ability info isn't available from the log (i.e. vanilla game). As such, show if allowed.
+	if not data.p.abilities[1] and PokemonData.canShowUnknownAbilities() then
+		data.p.abilities[1] = PokemonData.getAbilityId(pokemonID, 0)
+	end
+	if not data.p.abilities[2] and PokemonData.canShowUnknownAbilities() then
+		data.p.abilities[2] = PokemonData.getAbilityId(pokemonID, 1)
+	end
```
Fellshadow, line 645: Same as the other comment about types, these should probably be combined to either get both from log or both from memory

Fellshadow (commented):
Agree with Kurumas on the mixing of sources of data, it feels a little unclean though it's more of a minor thing ultimately

Did a quick test and it looks like this does resolve #549 when playing regularly, however turning open book mode on reproduced the issue as I suspected

So we would ideally like something to differentiate current vs previous log - could a variable set by the open log buttons and passed into the data helper (or set in a global it can access) work?

ironmon_tracker/data/DataHelper.lua
UTDZac: Good note, and I agree. Fixed as of https://github.com/besteon/Ironmon-Tracker/pull/550/commits/b03d936ddf476cdf6d5e530180596683cea3b5f6

AndreVolpi (approved):
Code-wise lgtm, have no setup to test atm though.

ironmon_tracker/data/DataHelper.lua
Fellshadow:
```lua
if not data.p.types[1] and PokemonData.canShowUnknownTypes() then
	data.p.types[1] = pokemonInternal.types[1]
	if not data.p.types[2] then
		data.p.types[2] = pokemonInternal.types[2]
	end
end
```
I don't think this changes the behaviour, the first type is still pulled from memory in this scenario whether the second type exists or not

for a combined all-or-nothing it'd be something more like this I think
```lua
if not data.p.types[1] and not data.p.types[2] and PokemonData.canShowUnknownTypes() then
	data.p.types[1] = pokemonInternal.types[1]
	data.p.types[2] = pokemonInternal.types[2]
end
```

AndreVolpi: On Zac's version or Fell's, the end result is the same for the current randomizers/romhacks, because there is no case where checking the second ability matters, if the first is empty, the second should always be empty, but the problem was when the second was empty and the first wasn't, which is quite common. Though Fell's version covers the case if a new randomizer somehow populates only the second ability.

Fellshadow: yeah good point, and I doubt a randomizer like that would happen since that would probably cause other problems lmao

Fellshadow (commented):
Current version lgtm, can change as per my most recent comment if you want but as kurumas pointed out probably doesn't matter too much

Main outstanding thing is differentiating current vs previous log, to avoid the issue occurring when playing open book - whether you want to tackle that as part of this PR or after it is up to you, I don't think too many people play open book but there's also no rush to get the fix out so may as well get a full fix in on the release

UTDZac:
I was in the process of making both of these changes, you two were too fast 😆 

Added a tracking variable that can confirm if the opened log is for the currently loaded game, as of https://github.com/besteon/Ironmon-Tracker/pull/550/commits/23066fbdbe7dce4e2ce67f0abc8ba3f65972b140

AndreVolpi (approved):

Fellshadow (approved):
From a quick test I couldn't seem to reproduce the issue both when open book was disabled and enabled

PR #551, opened 2026-09-07
Patch 9.4.0

### LINKED ISSUE #549: [Bug]: Incorrect info from memory reads sometimes displayed when viewing previous logs

UTDZac:
### Release Notes
- Added an option to choose how the Tracker's Language setting gets used: menus, game stuff, or both
- Fixed an issue where incorrect data was being displayed under certain conditions when viewing the Previous Log
- Fixed a trainer id lookup issue for some IronMON ROM hacks

Note: This does not resolve the strange TAR extraction issue that affects a rare set of users. No fix was found that works for that which also doesn't break things for the majority of others.

### DIFF
--- a/ironmon_tracker/Resources.lua
+++ b/ironmon_tracker/Resources.lua
@@ -69,7 +69,7 @@ function Resources.initialize()
 	Resources.sanitizeTable(Resources.Languages)
 
 	-- Load in default language data first, to fill in gaps in other language data
-	Resources.loadAndApplyLanguage(Resources.Default.Language)
+	Resources.applyAndLoadLanguage(Resources.Default.Language)
 	FileManager.copyTable(Resources.Data, Resources.Default.Data)
 
 	-- Define metatables
@@ -82,7 +82,7 @@ function Resources.initialize()
 		Resources.autoDetectForeignLanguage()
 	else
 		local userLanguageKey = Options["Language"] or Resources.Default.Language.Key
-		Resources.loadAndApplyLanguage(Resources.Languages[userLanguageKey])
+		Resources.applyAndLoadLanguage(Resources.Languages[userLanguageKey])
 	end
 
 	Resources.hasInitialized = true
@@ -105,7 +105,7 @@ function Resources.autoDetectForeignLanguage()
 end
 
 function Resources.changeLanguageSetting(language, forced)
-	local success = Resources.loadAndApplyLanguage(language)
+	local success = Resources.applyAndLoadLanguage(language)
 	if success or forced then
 		Resources.currentLanguage = language
 		Options["Language"] = language.Key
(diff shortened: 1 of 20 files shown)


ironmon_tracker/Resources.lua
```diff
@@ -159,10 +170,20 @@ function Resources.defineResourceCallbacks()
...
+		end
+		dataLoadHelper("Game", data)
+	end
 
 	-- Each screen is its own asset category of data
 	function ScreenResources(data)
+		-- Don't replace Tracker UI menus and text with language if setting is off
+		if not Options["Show menu text in language"] then
+			return
+		end
```
Fellshadow, line 186: The early returns here break things when the tracker is loaded (either initially or on loading new roms)

If you load the tracker, and toggle either of the "Show ... in language" settings **off** then reload the tracker (either by stopping and restarting the script or loading a different rom), the tracker crashes with an error in the lua console about nil resource data because these early returns result in not loading the associated resource data at all

here are the current errors I saw, though the issue is more than just these specific fields
game words turned off
```
NLua.Exceptions.LuaScriptException: ironmon_tracker\data\PokemonData.lua:182: attempt to index field 'Game' (a nil value)
```
menu text turned off
```
NLua.Exceptions.LuaScriptException: ironmon_tracker\data\PokemonData.lua:191: attempt to index local 'RPED' (a nil value)
```

Fellshadow (changes requested):
Code mostly lgtm, however the resource loading needs revised as turning either of the new settings off causes the tracker to crash on reload when I tested it

other changes seem fine to me from quick testing

Fellshadow:
just sliding in [efda7eb](https://github.com/besteon/Ironmon-Tracker/commit/efda7ebe5cd1ca30f45458d82d202bef7a65f995) while we've got something going to main, nothing tracker-related just updating the github issue template for bug reports

ironmon_tracker/Resources.lua
UTDZac: Woah. Good catch. I should have tried turning it off and on again after changing the setting. Thanks for the find.

Fixed in d904dd3547cbb2ee20a22aa2d18f672fe5972e65

Fellshadow (approved):
Fix appears to be working from what I tested, so all good now from what I can tell

PR #552, opened 2026-09-14
Fix Type effectiveness leaking type knowledge when types are randomized

SJTButler:
Hi, while trying to play around with re-implementing the Ironmon tracker on 3DS, I discovered an issue with the logic for the "Show Random Move/Type data" option whereby if you have a Pokemon rom set up with randomised types, while the types themselves are hidden, the weaknesses and resistances of the Pokemon aren't, which could allow someone to work out (or at least narrow down) the type of the Pokemon they are facing. This can be seen on the Pokemon Info screen (reached by tapping the Pokemon sprite when in a battle), and in the weaknesses screen (reached by tapping the type icon). 

I don't believe that this is a very common use case for the tracker (are there any Ironmon setups that require Pokemon types to be randomised?), so I can understand how this hasn't been spotted before.

I have tried to fix this by changing the places where Pokemon effectiveness is obtained so that, if the settings indicate that random types should be hidden, the effectiveness dictionaries will contain only a 2X weakness to PokemonData.Types.UNKNOWN, so the screens will show that icon (the same as the one showed when types are hidden). I have also added checks to see if the logs are Open, in which case the effectiveness will show as normal (like with types).

I don't know if I have fixed the problem in the best way - it might be better to centralise the "should type effectiveness be displayed" logic into one place, and have a single function to pull in the real or fake effectiveness dictionary based on that logic, so please feel free to send me away to have another go at it. Alternatively, if you would rather me leave this to you and open an Issue instead, please let me know.

Thanks for all the hard work on the Tracker!


UTDZac:
Hi @SJTButler , thanks for the issue report and the code. I'd like to look into it a bit more and figure out the root cause of this.

Would you be able to share step-by-step how I can reproduce this issue on my end? I'd like to see the issue for myself, as well as test it against your code changes (and further review them). Any screenshots would also be helpful

SJTButler:
Hi UTDZac - have opened an issue here https://github.com/besteon/Ironmon-Tracker/issues/553 with steps to reproduce and some screenshots - please let me know if there's any more info you want from me :)

